PR #40, opened 2023-10-24
[JAVAVSCODE-28] Include TestNG tests in the test explorer.

### LINKED ISSUE #28: vscode test explorer does not show TestNG tests
The gradle runner runs the tests fine, but the vscode test explorer only seems to show tests that are JUnit tests.

The project has both JUnit and TestNG tests.

The build.gradle has

test {
    useTestNG()
}

I realize this is a weird setup, but I would think the extension would use the gradle file as the SOT and not find some JUnit tests and then only show JUnit tests.

lahodaj:
Includes:
https://github.com/apache/netbeans/pull/6615
into the extension. Please wait with merging until the NetBeans PR is approved.

Thanks!

closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ In the VS Code command palette :
 * __Java: New Project...__ allows creation of new Maven or Gradle project 
 * __Java: New from Template...__ add various files to currently selected open project. Files are:
     * Java - broad selection of various predefined Java classes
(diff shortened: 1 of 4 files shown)


lahodaj:
The patch is included, as pointed out in https://github.com/lahodaj/javavscode/pull/2. README update may still be in order.

Achal1607 (approved):

PR #67, opened 2023-11-06
[JAVAVSCODE-22] Added support for using different jdk in each workspace and deprecated jdk.userdir configuration

Achal1607:
- Updated `jdk.jdkhome` to `machine-overridable` from `machine` so that different jdk binaries can be used in each workspace. Tested this configuration for remote SSH as well and it is working fine with it as well.
- Deprecated `jdk.userdir` configuration and default behaviour is now having separate `userdir` for each workspace and a global `userdir` if no workspace is open.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -85,7 +85,7 @@
 					],
 					"default": null,
 					"description": "Specifies JDK for the Oracle Visual Studio Code Extension",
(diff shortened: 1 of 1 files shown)


lahodaj (approved):
Looks OK to me.

arvindaprameya (approved):

PR #75, opened 2023-11-10
[JAVAVSCODE-16] Added option for running any Java project using context menu

### LINKED ISSUE #16: Feature request: convenience methods to run/debug app
It would be great if we could make it easier to run an app. Two things I would love to see:

* The ability to right-click on a Java file and select Run or Debug 
* A "play"-button in the gutter for classes and runnable methods (main methods + tests), similar to IntelliJ

Achal1607:
Added options in context menu for running and debugging any Java project. i.e. getting run java and debug java options when we right click on the java file, also we get these options when we right click on the code editor. 

Closes #16 

Image of how options look when we right click on a Java file in workspace

image

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -467,6 +467,16 @@
 					"command": "jdk.java.goto.super.implementation",
 					"when": "nbJdkReady && editorLangId == java && editorTextFocus",
 					"group": "navigation@100"
+				},
+				{
+					"command": "jdk.project.run",
+					"when": "nbJdkReady && editorLangId == java && resourceExtname == .java",
+					"group": "javadebug@1"
+				},
+				{
+					"command": "jdk.project.debug",
+					"when": "nbJdkReady && editorLangId == java && resourceExtname == .java",
+					"group": "javadebug@2"
 				}
 			],
 			"explorer/context": [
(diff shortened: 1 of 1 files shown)


vscode/package.json
```diff
@@ -467,13 +467,33 @@
 					"command": "jdk.java.goto.super.implementation",
 					"when": "nbJdkReady && editorLangId == java && editorTextFocus",
 					"group": "navigation@100"
+				},
+				{
+					"command": "jdk.project.run",
+					"when": "nbJdkReady && editorLangId == java && resourceExtname == .java",
```
lahodaj, line 473: Is there a reason to use both `editorLangId == java` and `resourceExtname == .java`? I would assume one should be enough, but don't mind too much.

lahodaj (approved):
Looks reasonable. One comment for consideration inline.

vscode/package.json
Achal1607: Yes only `resourceExtname == .java` would be fine I guess but to be completely confident I have added both flags to check.

lahodaj: Ok. I don't object to this.

arvindaprameya:
Looks good to me

PR #81, opened 2023-11-16
[JAVAVSCODE-78] Added checksum match for jdk downloader

### LINKED ISSUE #78: Implement checksum match in JDK Downloader
Implement checksum match for JDK downloader as a step of better security.

Achal1607:
Added checksum match after downloading JDK binaries in JDK downloader feature

closes #78

### DIFF
--- a/vscode/src/jdkDownloader.ts
+++ b/vscode/src/jdkDownloader.ts
@@ -21,11 +21,26 @@ import axios from 'axios';
 import * as https from 'https';
 import * as child_process from 'child_process';
 import * as vscode from 'vscode';
+import * as crypto from 'crypto';
 import { JDK_RELEASES_TRACK_URL, OPEN_JDK_VERSION_DOWNLOAD_LINKS, ORACLE_JDK_BASE_DOWNLOAD_URL, ORACLE_JDK_VERSION_FALLBACK_DOWNLOAD_VERSIONS } from './constants';
 import { handleLog } from './extension';
 
 let customView: vscode.WebviewPanel;
 let logger: vscode.OutputChannel;
+
+export const calculateChecksum = (filePath: string) => {
+  try {
+    const ALGORITHM = 'sha256';
+    const stream = fs.readFileSync(filePath);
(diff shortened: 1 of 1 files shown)


vscode/src/jdkDownloader.ts
```diff
@@ -173,8 +188,16 @@ export function JDKDownloader(JDKType: string, osType: string, osArchitecture: s
...
-      await extractJDK(path.join(targetDirectory, fileName), installationPath, JDKVersion, osType);
+      const checkSumObtained = calculateChecksum(filePath);
+      const checkSumExpected = (await axios.get(`${downloadUrl}.sha256`)).data;
+      
+      if (checkSumExpected === checkSumObtained) {
+        await extractJDK(filePath, installationPath, JDKVersion, osType);
+      }
+      else {
+        handleLog(logger, "Checksums don't match!");
+        throw new Error("Something went wrong");
```
lahodaj, line 199: I tried what happens when the checksum fails, by hardcoding:
```
handleLog(logger, "Checksums don't match!");
throw new Error("Something went wrong");
```
after `showInformationMessage`. And it does not seem very user-friendly to me: it will show "... download completed!", and that's it. I think I would move the "... download completed!" message to the branch where the checksum passes, and would use something like `showErrorMessage` to report the problem. Possibly something along the lines of "${JDKType} ${JDKVersion} for {$osType} download failed, wrong checksum."

The `throws new Error` also seems a bit harsh - when the message is show, maybe we could simply not do anything else?

lahodaj (changes requested):
Some comments for consideration inline

vscode/src/jdkDownloader.ts
Achal1607: Yes, it didn't seem user friendly, I have updated the code according to the comment given by you. Thanks.

lahodaj (approved):
Looks reasonable to me.

arvindaprameya (approved):
LGTM

PR #82, opened 2023-11-16
[JAVAVSCODE-80] Renamed downloaded jdk binaries so that it is not replaced by the already present binaries

### LINKED ISSUE #80: JDK downloader: Allow download of Oracle JDK and openJDK binaries in the same directory
Currently, installing Oracle JDK and openJDK in the same directory is not possible using JDK downloader feature of extension. So, to allowing users to install both JDKs in the same directory would be good to have.

Achal1607:
Renamed downloaded extracted jdk binaries so that even if openJDK and oracle JDK binaries are downloaded in the same directory they are not replaced by each other.

closes #80

### DIFF
--- a/vscode/src/jdkDownloader.ts
+++ b/vscode/src/jdkDownloader.ts
@@ -190,39 +190,44 @@ export function JDKDownloader(JDKType: string, osType: string, osArchitecture: s
 }
 
 export async function extractJDK(jdkTarballPath: string, extractionTarget: string, jdkVersion: string, osType: string, jdkType: string): Promise<void> {
+  // Extract jdk binaries in a temp folder
   const downloadedDir = path.join(__dirname, 'jdk_downloads');
-  const extractCommand = `tar -xzf "${jdkTarballPath}" -C "${downloadedDir}"`;
+  const extractCommand = `tar -xzf "${jdkTarballPath}" -C ${downloadedDir}`;
+  const tempDirName = `jdk-${jdkVersion}${osType === 'macOS' ? '.jdk' : ''}`;
+  const tempDirectoryPath = path.join(downloadedDir, tempDirName);
+  let newDirectoryPath: string | null = null;
 
-  const oldDirName = `jdk-${jdkVersion}`;
-  const oldDirectoryPath = path.join(downloadedDir, osType === 'macOS' ? oldDirName + '.jdk' : oldDirName);
-  if (fs.existsSync(oldDirectoryPath)) {
-    await fs.promises.rmdir(oldDirectoryPath, { recursive: true });
-  }
   child_process.exec(extractCommand, async (error) => {
     if (error) {
       vscode.window.showErrorMessage('Error: ' + error);
     } else {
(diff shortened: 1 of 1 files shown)


vscode/src/jdkDownloader.ts
```diff
@@ -220,6 +220,19 @@ export async function extractJDK(jdkTarballPath: string, extractionTarget: strin
...
 }
 
+const handleJdkPaths = async (directoryName: string, parentPath: string, osType: string): Promise<string> => {
+  let name = directoryName;
+  if (osType === 'macOS') {
+    name = `${directoryName}.jdk`;
+  }
+  const directoryPath = path.join(parentPath, name);
+  if (fs.existsSync(directoryPath)) {
+    await fs.promises.rmdir(directoryPath, { recursive: true });
```
lahodaj, line 230: I wonder if there should be some warning to (or confirmation from) the user before deleting the directory?

Achal1607: Yes it sounds good to have confirmation, I have added it.

vscode/src/jdkDownloader.ts
```diff
@@ -189,25 +189,33 @@ export function JDKDownloader(JDKType: string, osType: string, osArchitecture: s
...
-
-  const extractCommand = `tar -xzf "${jdkTarballPath}" -C "${extractionTarget}"`;
+export async function extractJDK(jdkTarballPath: string, extractionTarget: string, jdkVersion: string, osType: string, jdkType: string): Promise<void> {
+  const downloadedDir = path.join(__dirname, 'jdk_downloads');
+  const extractCommand = `tar -xzf "${jdkTarballPath}" -C "${downloadedDir}"`;
 
+  const oldDirName = `jdk-${jdkVersion}`;
+  const oldDirectoryPath = path.join(downloadedDir, osType === 'macOS' ? oldDirName + '.jdk' : oldDirName);
+  if (fs.existsSync(oldDirectoryPath)) {
+    await fs.promises.rmdir(oldDirectoryPath, { recursive: true });
```
lahodaj, line 199: I think this has the same problem with deleting pre-existing directories without notification. I wonder: is it possible to create a temporary directy, unpack to it, and then move/rename to the correct location? I think it is OK to remove directories we create (e.g. when something goes wrong), but removing directories that exist without explicit consent seems wrong to me.

Achal1607: I have updated the code and following is the flow :
- Oracle JDK or OpenJDK `tar.gz` file will be downloaded
- It would be extracted in a temp folder
- If directory with same name is present in the user selected download location then ask user if they want to delete it or not? 
     - If user agrees for deleting the directory then delete it and move the temp directory to the user selected location with the updated name
     - If user disagrees for deleting the directory then do nothing
- Cleanup the temp folder and the `tar.gz` file

lahodaj (dismissed):
Seems OK.

arvindaprameya (dismissed):
LGTM

arvindaprameya (approved):
LGTM

PR #108, opened 2024-01-25
added support for gradle 8.5

### LINKED ISSUE #29: vscode reports a need to upgrade Gradle to an earlier version for JDK-21
I am using JDK-21 and Gradle 8.4, but when I open the project it tells me that the current project is not supported by Gradle 8.4 and I should install 8.3-rc-1.

It seems to work of with 8.4.

image

Achal1607:
Patch for gradle 8.5 added so that the misleading pop-up doesn't appear.
closes #29

### DIFF
--- a/patches/6807.diff
+++ b/patches/6807.diff
@@ -1,0 +1,37 @@
+diff --git a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
+index 1cc156d0f622..c90316077493 100644
+--- a/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
++++ b/extide/gradle/src/org/netbeans/modules/gradle/api/execute/GradleDistributionManager.java
+@@ -100,6 +100,7 @@ public final class GradleDistributionManager {
+         GradleVersion.version("7.5"), // JDK-18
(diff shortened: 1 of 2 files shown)


patches/6807.diff
```diff
@@ -0,0 +1,37 @@
...
+          * @return <code>true</code> if this version is supported with that JDK.
+          */
+         public boolean isCompatibleWithJava(int jdkMajorVersion) {
+-            return jdkMajorVersion <= lastSupportedJava();
++            
++            GradleVersion lastKnown = JDK_COMPAT[JDK_COMPAT.length - 1];
++            // Optimistic bias, if the GradleVersion is newer than the last NB
++            // knows, we say it's compatible with any JDK
++            return lastKnown.compareTo(version.getBaseVersion()) < 0 
++                    || jdkMajorVersion <= lastSupportedJava();
```
sundararajana, line 34: can we make use of  Gradle compatibility table given here https://docs.gradle.org/current/userguide/compatibility.html ?

Achal1607: Yes maybe I don't know for sure, but currently it is manually handled in netbeans for every release. This patch is taken from https://github.com/apache/netbeans/pull/6807

capricube:
who is able to give the final approval? The error message when using the Java Platform Support extension is very annoying.

Achal1607:
> who is able to give the final approval? The error message when using the Java Platform Support extension is very annoying.

The change is available in NB-21 and it would be released in next version of the extension

Achal1607:
Available in NB-21. So, closing this PR.

PR #116, opened 2024-02-21
Option to add extra args in vsce package script

Achal1607:
Option to add extra args in vsce package script like `--pre-release` etc.


lahodaj (approved):
Looks good to me.

arvindaprameya (approved):

Issue #30, opened 2023-10-22
Renaming the class containing the main method does not update pom.xml


tofflos:
Create a new Maven project using Ctrl-Shift-P Java New Project.
Check that it's possible to Run the project.
Check that pom.xml contains an entry similar to `<exec.mainClass>com.example.Main</exec.mainClass>` in the properties section.

Open the class containing the main method, select the class name, press F2, and rename it.
Note that the file has been renamed.
Note that it's no longer possible to Run the project.
Note that the pom.xml entry `<exec.mainClass>com.example.Main</exec.mainClass>` hasn't been updated.

Achal1607:
FYI Raised a PR on netbeans side:
https://github.com/apache/netbeans/pull/7104

PR #120, opened 2024-02-24
added maven refactoring module from the netbeans side

Achal1607:
This PR is for enabling refactoring of pom.xml as well on renaming class name,

For more info: #30 
PR netbeans: https://github.com/apache/netbeans/pull/7104


lahodaj (approved):
Looks good!

PR #122, opened 2024-03-13
upgraded axios to 1.6.7 from 1.6.0

Achal1607:
Upgraded axios to 1.6.7
indirectly upgrading follow-redirects to 1.15.5 from 1.15.0(which had vulnerablility)


arvindaprameya (approved):
LGTM, thank you

Issue #110, opened 2024-01-29
Extension fails to start


anthonyvdotbe:
When I open VS Code, I see the following errors and the extension fails to start. (And yes, C:\dev\jdk-21.0.2 contains a valid JDK installation.)

![image](https://github.com/oracle/javavscode/assets/15714253/d14c3791-debd-46ac-92e5-cabe48957578)

extension: 1.0.0
VS Code: 1.85.2 (8b3775030ed1a69b13e4f4c628c612102e30a681)
OS: Windows_NT x64 10.0.22631

Log:
```
Initiating server activation
Request to kill LSP server.
Cannot kill: current process is None
Launching Oracle Java SE Language Server with C:\dev\jdk-21.0.2 and userdir c:\Users\antho\AppData\Roaming\Code\User\workspaceStorage\a3a1ac302a43d9d618aae91e36299438\Oracle.oracle-java
LSP server launching: 13732
LSP server user directory: c:\Users\antho\AppData\Roaming\Code\User\workspaceStorage\a3a1ac302a43d9d618aae91e36299438\Oracle.oracle-java
-------------------------------------------------------------------------------
>Log Session: Monday, January 29, 2024, 11:18:18 AM Central European Standard Time
>System Info: 
  Product Version         = vsc dev-0cdb2f58d52f66abcc50da68aadc9c93bd94c2f8
  Operating System        = Windows 11 version 10.0 running on amd64
  Java; VM; Vendor        = 21.0.2; Java HotSpot(TM) 64-Bit Server VM 21.0.2+13-LTS-58; Oracle Corporation
  Runtime                 = Java(TM) SE Runtime Environment 21.0.2+13-LTS-58
  Java Home               = C:\dev\jdk-21.0.2
  System Locale; Encoding = en_US (nbcode); UTF-8
  Home Directory          = C:\Users\antho
  Current Directory       = c:\Users\antho\.vscode\extensions\oracle.oracle-java-1.0.0\nbcode
  User Directory          = C:\Users\antho\AppData\Roaming\Code\User\workspaceStorage\a3a1ac302a43d9d618aae91e36299438\Oracle.oracle-java\userdir
  Cache Directory         = C:\Users\antho\AppData\Roaming\Code\User\workspaceStorage\a3a1ac302a43d9d618aae91e36299438\Oracle.oracle-java\userdir\var\cache
  Installation            = C:\Users\antho\.vscode\extensions\oracle.oracle-java-1.0.0\nbcode\platform
[...]

PR #124, opened 2024-03-15
Added option to delete workspace cache with respect to oracle java extension 

Achal1607:
As seen in issue #110, sometimes cache issue can occur and user needs to delete userdir as part of troubleshooting step for that workspace. So, added a command to delete extension cache for a particular workspace.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -445,6 +445,10 @@
 				"command": "jdk.goto.test",
 				"title": "Go To Test/Tested class...",
 				"category": "Java"
+			},
+			{
+				"command": "jdk.delete.cache",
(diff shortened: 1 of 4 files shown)


lahodaj (approved):
Looks OK to me.

vscode/package.json
```diff
@@ -445,6 +445,10 @@
 				"command": "jdk.goto.test",
 				"title": "Go To Test/Tested class...",
 				"category": "Java"
+			},
+			{
+				"command": "jdk.delete.cache",
+				"title": "Delete oracle java extension cache for this workspace"
```
StanFromTraction, line 451: Oracle and Java should be capitalised

PR #129, opened 2024-03-19
removed db support from frontend as well

Achal1607:
- Removed DB unused code from frontend
- Fixed `remove-db.diff` patch it was not applying on nb-21


arvindaprameya (approved):
Approved

PR #130, opened 2024-03-20
Adding an advanced option to disable the embedded nb-javac, hence use javac from the JDK.

lahodaj:
This should allow easier experiments with early access JDK builds - those can be specified in `jdk.jdkhome`, and nb-javac can be disabled, so that the javac from the early access build is used by the extension.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -194,6 +194,11 @@
 					"type": "string",
 					"default": "",
 					"description": "Working directory"
+				},
+				"jdk.advanced.disable.nbjavac": {
+					"type": "boolean",
+					"default": false,
+					"description": "Advanced option: disable nb-javac library, javac from the selected JDK will be used. The selected JDK must be at least JDK 22."
 				}
 			}
 		},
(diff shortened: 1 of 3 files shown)


arvindaprameya (approved):
LGTM, thank you

Achal1607 (commented):
I tried to validate it and had one question, does installed label here means "module not enabled"?
Screenshot 2024-03-21 at 10 22 31 AM

lahodaj:
Yes, Installed means present, but not enabled:
https://github.com/apache/netbeans/blob/e35c66fa121a4c76f13e4a2474cba7d593fec08a/platform/autoupdate.cli/src/org/netbeans/modules/autoupdate/cli/Status.java#L82

The best way to try this patch may be to take JDK 23, and try pattern matching over a primitive type and enable preview. I.e. something like:
```
Object o = 1;
boolean b = o instanceof int i;
```

Should produce an error without the JDK 23 javac, and should be OK with JDK 23 javac and preview enabled.

Thanks!

Achal1607 (approved):
Validated using some examples from this JEP: https://openjdk.org/jeps/455. It is working fine.

PR #133, opened 2024-03-21
Added JDK 22 GA download option and refactored code a bit

### LINKED ISSUE #131: JDK 22 GA download option addition in JDK Downloader
JDK 22 GA is now released so need to add an option to download JDK 22

Achal1607:
- Added JDK 22 GA download option fr both Oracle and Oracle OpenJDK.
- Since on every JDK release we need to update the download labels it gets problematic. So, removed specific version numbers and switched to displaying only JDK family version i.e. 21,17 or 22.
- Also, removed version fetching from `releases.json` file as it was not working.

Before
image

Now 
image

jdkDownloaderManagement.json is kept 
```
{  
  oracleJdk: {
             "version":{ baseDownloaderUrl: "downloadUrl"}
  }, 
  openJdk: {
             "version":{ baseDownloaderUrl: "downloadUrl"}
  }
}
```
So, that in future if some more configuration for a particular JDK is wanted then we can add here without disturbing existing structure.

closes #131


arvindaprameya (approved):
LGTM, thankyou

arvindaprameya (approved):
LGTM, thank you!!

PR #134, opened 2024-03-25
Removed glob usage from JDK Downloader

Achal1607:
Removed `glob` code and wrote alternate code using `fs` lib which is already present in node js.


arvindaprameya (approved):
LGTM

PR #179, opened 2024-06-18
Performance tests

Achal1607:
As a starting step added performance tests to measure how much time it takes to open OpenJDK project.


vscode/src/test/suite/performance/open-jdk.test.ts
```diff
@@ -0,0 +1,135 @@
...
+const checkSymbolsResolved = (path: string): boolean => {
+    let d = languages.getDiagnostics(Uri.file(path));
+    const filterErrorsList = d.filter(el => el.severity == 0);
+    console.log("Filtered Errors List Length: " + filterErrorsList.length);
+    return filterErrorsList.length == 0;
+}
+
+const checkExtensionLoaded = (path: string): boolean => {
+    let d = languages.getDiagnostics(Uri.file(path));
+    return d.length != 0;
```
sid-srini, line 20: Do add a comment on `constants.OPENJDK_CHECK_FILES_RESOLVES` that the files chosen must have some warnings or other diagnostic information/hints generated. Otherwise, this method would fail to detect that the extension loaded.

vscode/src/test/suite/performance/open-jdk.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import { assertWorkspace, openFile, runShellCommand, waitCommandsReady } from '../../testutils';
+import { OPENJDK_CHECK_FILES_RESOLVES } from '../../constants';
+
+let lastFileSize = 0;
+let count = 0;
+const checkSymbolsResolved = (path: string): boolean => {
+    let d = languages.getDiagnostics(Uri.file(path));
+    const filterErrorsList = d.filter(el => el.severity == 0);
+    console.log("Filtered Errors List Length: " + filterErrorsList.length);
+    return filterErrorsList.length == 0;
```
sid-srini, line 15: Perhaps the return should check that there are some diagnostics available and then no errors are reported. 
```js
    return d.length > 0 && filterErrorsList.length == 0;
```

For the first file, this would be done by the separate invocation of `checkExtensionLoaded` and so it may not have any impact when there is only a single file in the check list.

sid-srini (changes requested):

PR #192, opened 2024-07-05
pre-commit hook to check urls in package-lock.json

Achal1607:
pre-commit hook added to check if internal artifactory link is not present in package-lock.json.

### DIFF
--- a/vscode/esbuild.js
+++ b/vscode/esbuild.js
@@ -32,7 +32,7 @@ const watchConfig = {
   };
 
 const NON_NPM_ARTIFACTORY = new RegExp(
-  String.raw`"resolved": "(?!https://registry.npmjs.org).*"`,
+  String.raw`(")resolved\1\s*:\s*(")http[s]*://(?!registry.npmjs.org)[^"']+\2`,
   "g"
 );
 
--- a/vscode/esbuild.js
+++ b/vscode/esbuild.js
@@ -1,4 +1,6 @@
 const { build } = require("esbuild");
+const fs = require('fs');
+const path = require('path');
 
 const baseConfig = {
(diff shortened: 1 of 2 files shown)


vscode/esbuild.js
```diff
@@ -28,7 +30,21 @@ const watchConfig = {
       },
     },
   };
-  
+
+const NON_NPM_ARTIFACTORY = new RegExp(
+  String.raw`"resolved": "(?!https://registry.npmjs.org).*"`,
```
sid-srini, line 35: Minor regex suggestions for being whitespaces and quotes agnostic.
```re
("|')resolved\1\s*:\s*("|')http[s]*://(?!registry.npmjs.org)[^"']+\2
```
Do test the suggested regex for functional correctness if you choose to use it.

sid-srini (changes requested):

vscode/esbuild.js
Achal1607: Since it is searching for the pattern in `package-lock.json`. It would have double quotes only, correct?
So, should I change `("|')` to `(")`

sid-srini: Yes, you're right. The package-lock.json is expected to be from a code generator, not created manually. So, it is expected to stick to the JSON standard of double-quotes. 
In general, however, when reading any input data, it is necessary to be more lenient. In that case, it would have been more appropriate to use a JSON parser and test the values.
So please go ahead with the simpler regex like:
```re
"resolved"\s*:\s*"http[s]*://(?!registry.npmjs.org)[^"]+"
```
Thanks.

Achal1607:
@sid-srini can you please review it again?

PR #211, opened 2024-07-25
[JAVAVSCODE #199] Quick Fix actions are unable to edit runConfig options in global settings for non-workspace opened Java files

### LINKED ISSUE #199: Quick Fix actions are unable to edit `runConfig` options in global settings for non-workspace opened Java files
# Summary
When a Java file is opened directly in the VS Code editor, ***without opening a folder or a workspace file***, then the Quick Fix actions, such as *Enable Preview Feature*, are unable to save the `runConfig` options in the global settings.

## Steps to Reproduce
0. Do not have VS Code open with a folder or workspace prior to the following steps.
    - Install the Oracle Java extension (I've tried with v. 22.0.1)
1. Open a Java file containing only a `main` method directly in VS Code using the *Open* menu action or from Finder/Explorer.
    - Alternatively, create a *New* Java file and write a `main` method.
    - This would require the preview feature of Java 22 to be enabled.
2. Accept to *trust* the file or folder request from VS Code, if prompted.
3. Notice that after analysis by the Java extension, the editor shows:
    - a *Problem* that: `implicitly declared classes are a preview feature and are disabled by default`,
    - with a *Quick Fix* to ***Enable Preview Feature***
    - Screenshot 2024-07-23 at 11 01 58 AM
    - Screenshot 2024-07-23 at 11 03 05 AM
4. However, on selecting the *Quick Fix* action, nothing seems to happen.
    - None of the global extension `runConfig` settings are updated.

## Impact
The Quick Fix actions which require modifying the `runConfig` global extension settings do not seem to run and validation of the Java application continues to fail.

# Workaround
Manually edit the global settings with the required changes.
Example: `"jdk.runConfig.vmOptions": "--source 22 --enable-preview"`

# Fix Request
The `runConfig` global settings should get modified by the *Quick Fix* actions, such as, *Enable Preview Feature* even for non-workspace opened Java files. 
This will help in quickly trying out new preview features in single-file Java code without setting up a separate test folder or workspace, or, manually edit settings which only advanced users may do.

sid-srini:
Fixed extension.ts `UpdateConfigurationRequest` handler to update the global `WorkspaceConfiguration` for non-workspace opened files.

1. Checked for non-workspace opened files by testing if `vscode.workspace.workspaceFile` is undefined/null.
2. Invoked `WorkspaceConfiguration.update()` with `configurationTarget = `
    - `true`: for non-workspace files;
    - `null`: otherwise.
4. Added a try-catch to log errors in update and prevent downstream failures.
5. Avoided an unnecessary `await` and used the `Thenable` chain of `update()`.

Closes #199

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -1101,21 +1101,19 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
             });
             c.onRequest(UpdateConfigurationRequest.type, async (param) => {
                 handleLog(log, "Received config update: " + param.section + "." + param.key + "=" + param.value);
-                if (vscode.workspace) {
-                    let wsFile: vscode.Uri | undefined = vscode.workspace.workspaceFile;
-                    let wsConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(param.section);
-                    if (wsConfig) {
-                        try {
-                            wsConfig.update(param.key, param.value, wsFile ? null : true)
-                                .then(() => {
-                                    handleLog(log, "Updated configuration: " + param.section + "." + param.key + "=" + param.value + "; in: " + (wsFile ? wsFile.toString() : "Global"));
-                                })
-                                .then(() => {
-                                    runConfigurationUpdateAll();
-                                });
-                        } catch (err) {
-                            handleLog(log, "Failed to update configuration. Reason: " + (typeof err === "string" ? err : err instanceof Error ? err.message : "error"));
-                        }
+                let wsFile: vscode.Uri | undefined = vscode.workspace.workspaceFile;
+                let wsConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration(param.section);
+                if (wsConfig) {
+                    try {
+                        wsConfig.update(param.key, param.value, wsFile ? null : true)
+                            .then(() => {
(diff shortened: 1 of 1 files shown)


vscode/src/extension.ts
```diff
@@ -1100,8 +1100,24 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
                 return selected ? Array.isArray(selected) ? selected : [selected] : undefined;
             });
             c.onRequest(UpdateConfigurationRequest.type, async (param) => {
-                await vscode.workspace.getConfiguration(param.section).update(param.key, param.value);
-                runConfigurationUpdateAll();
+                handleLog(log, "Received config update: " + param.section + "." + param.key + "=" + param.value);
+                if (vscode.workspace) {
```
lahodaj, line 1104: Is there a case where `vscode.workspace` could be `undefined`? If not, I'd drop the if.

lahodaj (commented):
One comment inline.

vscode/src/extension.ts
sid-srini: Thanks @lahodaj for pointing this out. I've removed the `if` check since it will always be `true`.

- `vscode.workspace` is a VSCode API [namespace](https://code.visualstudio.com/api/references/vscode-api#workspace).

lahodaj (approved):
Looks good to me.

Issue #96, opened 2023-12-21
No support for @SupressWarnings


psaraviaq:
Unused variables are still underlined even when SupressWarnings is applied. This did not happen when working with the RedHat extension.

```
@SuppressWarnings("unused")
public class Abstraction {

    public static void main(String[] args) {
        Object object = new Object();
    }
}
```

![image](https://github.com/oracle/javavscode/assets/102195797/cda5706c-9a5a-48af-9737-fb24f0061ad5)
![image](https://github.com/oracle/javavscode/assets/102195797/9d5143fc-ef70-4ab9-8843-de92e320b981)

pbodnar:
Findings from my testing:

1. `@SuppressWarnings("unused")` really is ignored, while others, like `@SuppressWarnings("static-access")` are correctly honored.
2. Unlike the Red Hat extension, this one doesn't seem to supply Quick Fix for adding/extending `@SuppressWarnings` for any type of suppressible problem.

Achal1607:
PR raised on Netbeans side to fix this issue: https://github.com/apache/netbeans/pull/7548

Achal1607:
PR merged would be available in NB-23

PR #227, opened 2024-08-06
Backporting NetBeans 23 patches

### LINKED ISSUE #214: Error in alt + insert shortcut
command 'nbls.java.generate.getter.setter' not found

image

Achal1607:
Backporting some NB-23 patches to fix issues.
closes #214 and #96


patches/7584.diff
lahodaj, line 1: Two comments here: a) there's a typo in the file name, it should be 7548 (https://github.com/apache/netbeans/pull/7548). b) the patch here is much shorter than the one in the PR, but maybe that's intentional removing the sub-fixes?

Achal1607: a) I will fix the typo. Again silly mistake from my end sorry.
b) Yes it is intentional since this release would only contain bug fix so added that part only.

lahodaj (dismissed):
Looks OK to me.

sid-srini (approved):
LGTM

PR #231, opened 2024-08-09
Revert "Backporting NetBeans 23 patches"

Achal1607:
7548 would be coming in NB-23.
 
Reverts oracle/javavscode#227


Achal1607:
Not needed

PR #220, opened 2024-07-31
Upgrade Gradle tooling to support JDK-22

Achal1607:
Backporting NB-23 patch for upgrading Gradle to support JDK-22
https://github.com/apache/netbeans/pull/7583

closes #190

PR #232, opened 2024-08-09
NB-23 patches compatible with Java source version 1.8

Achal1607:
Patches that were backported from NB-23 in PRs: #220 and #227 made compatible to work with Java source version 1.8


lahodaj (approved):
Looks good, thanks!

PR #234, opened 2024-08-13
Ant build script reorganization for patches and clean-netbeans

sid-srini:
1. Placed patch files in separate lines to ease merges and reduce conflicts.

2. Added string parsing and trimming of the patches list to allow for placing patch files in separate lines.
    - Also added a check for empty patch list when applying or unapplying the patches.

3. Added a "clean-netbeans" target, similar to the "build-netbeans" target, to help in cleaning just the netbeans build artifacts.


Achal1607 (approved):
LGTM. Thanks for updating the build file

PR #236, opened 2024-08-13
[JVSC #199] Backport NetBeans 24 patch 7610

sid-srini:
Backporting the NetBeans 24 patch [#7610](https://github.com/apache/netbeans/pull/7610) and adding it to the patches list in build.xml.

- This brings the fix for getting config options for a single source file to the NB 22 branch.
- This allows the fix for #199 to demonstrate its effect.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -44,6 +44,7 @@
             patches/7548_source-1.8.diff
             patches/7583_source-1.8.diff
             patches/7621.diff 
+            patches/7610.diff
             patches/mvn-sh.diff
             patches/generate-dependencies.diff
             patches/rename-debugger.diff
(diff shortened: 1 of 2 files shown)


build.xml
```diff
@@ -44,6 +44,7 @@
             patches/7548_source-1.8.diff
             patches/7583_source-1.8.diff
             patches/7621.diff 
+            patches/7610.diff
```
Achal1607, line 47: While applying patches we generally try to maintain the patches in ascending order of PR numbers, so if possible can you please place this patch before `7621.diff`, it wouldn't make any difference in this case so it's up to you.

sid-srini: Ok. Thanks @Achal1607 for letting me know. I assumed that it was in ascending order of PR approval. I'll update the file. Thanks.

sid-srini: Done.

Achal1607 (approved):

Issue #194, opened 2024-07-15
Formatter doesn't support `{@return foo}`


anthonyvdotbe:
The formatter messes up Javadoc like `{@return foo}`.

For example formatting this:

```java
    /**
     * {@return foo}
     */
    String bar() {
        return null;
    }
```

Actual:

```java
    /**
     * {
     *
     * @return foo}
     */
    String bar() {
        return null;
    }
```

Expected:

```java
    /**
     * {@return foo}
     */
    String bar() {
        return null;
    }
```

naren2605:
create a [issue](https://github.com/apache/netbeans/issues/7640) in netbeans related to this issue

naren2605:
### verification of fix
before formatting bar method:
<img width="606" alt="image" src="https://github.com/user-attachments/assets/4bf63d4c-87c9-428a-be39-000f5b9d3545">



after formatting bar method:

<img width="1019" alt="image" src="https://github.com/user-attachments/assets/26164293-2d77-4542-920b-3c92bb765e22">

PR #237, opened 2024-08-14
[JAVAVSCODE #194] applying netbeans patch(pr-7641) - Formatter doesnt support {@return foo}

naren2605:
backporting commits from nb pr -https://github.com/apache/netbeans/pull/7641


build.xml
```diff
@@ -43,7 +43,8 @@
             patches/7491-preliminary.diff
             patches/7548_source-1.8.diff
             patches/7583_source-1.8.diff
-            patches/7621.diff 
+            patches/7621.diff
```
sid-srini, line 46: See if you can avoid the minor whitespace change on this line so that merge conflicts are avoided. Maybe in a later commit we can remove the extra whitespace as it will not have any conflicts then.

sid-srini (approved):
Looks good to me. Thanks.

sid-srini (approved):

PR #242, opened 2024-08-19
Replace axios with Node provided modules

Achal1607:
This PR aims to remove the axios module and instead use 'https' package provided by Node itself.

### DIFF
--- a/vscode/src/jdkDownloader.ts
+++ b/vscode/src/jdkDownloader.ts
@@ -17,7 +17,6 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import axios from 'axios';
 import * as https from 'https';
 import * as child_process from 'child_process';
 import * as vscode from 'vscode';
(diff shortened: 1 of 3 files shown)


vscode/src/jdkDownloader.ts
```diff
@@ -29,6 +29,22 @@ import { promisify } from 'util';
 let customView: vscode.WebviewPanel;
 let logger: vscode.OutputChannel;
 
+export const getSha256Checksum = async (url: string) => {
+  return new Promise((resolve, reject) => {
+      https.get(`${url}.sha256`, res => {
+          let data = '';
```
sid-srini, line 35: Perhaps you may add the check for the response status code 200 before attempting to read the data. Similar to the actual file download itself in the function `JDKDownloader`.

sid-srini (changes requested):
This PR would be very useful in removing an extra third-party library dependency. 
Please do remove the `axios` imports from *jdkDownloader.ts* and the dependency from *package.json* and *package-lock.json*. Thanks.

vscode/src/jdkDownloader.ts
```diff
@@ -28,9 +27,28 @@ import { promisify } from 'util';
 
 let customView: vscode.WebviewPanel;
 let logger: vscode.OutputChannel;
+const ALGORITHM = 'sha256';
+
+export const getSha256Checksum = async (url: string) => {
+  return new Promise((resolve, reject) => {
+      https.get(`${url}.${ALGORITHM}`, res => {
+          if (res.statusCode !== 200) {
+            return `HTTP Error ${res.statusCode} - ${res.statusMessage}`;
```
sid-srini, line 36: Should it throw an error here with the message string being returned here?
That way the rejection handler could appropriately notify the user.

Achal1607:
Closing since changes done in #248

PR #243, opened 2024-08-19
XML file formatting test added

Achal1607:
Added a new test case for checking formatting of xml file.


sid-srini (approved):
Looks fine to me. Thanks.

PR #244, opened 2024-08-22
Support to set Project level JDK

Achal1607:
- The ability to use a separate JDK for user projects/workspaces, distinct from the JDK used by the Oracle Java extension, has been added.
- Netbeans 23 onwards requires java version >=17, whereas user projects can run on java version >= 8.
- `jdk.project.jdkhome` new setting added for project level JDK, it would default to `jdk.jdkhome` setting.
- Updated README


sid-srini (approved):
Looks good to me. Thanks @Achal1607 and @lahodaj.

PR #245, opened 2024-08-26
Fix for delete cache error on windows

### LINKED ISSUE #221: "Delete Oracle Java extension cache for this workspace" fails with an error
Attempting to execute the command says:

image

which seems logical. Red Hat's extension has a similar command and restarts VS Code, most likely to avoid issues like this.

shivam71:
Killed the nb process and the stopped the LSP client so that no process is using the cache  before deleting the directory .
Closed #221

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -510,28 +510,28 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
 
     context.subscriptions.push(vscode.commands.registerCommand(COMMAND_PREFIX + ".delete.cache", async () => {
         const storagePath = context.storageUri?.fsPath;
-        if(!storagePath){
+        if (!storagePath) {
             vscode.window.showErrorMessage('Cannot find workspace path');
             return;
         }
-        
+
         const userDir = path.join(storagePath, "userdir");
         if (userDir && fs.existsSync(userDir)) {
             const confirmation = await vscode.window.showInformationMessage('Are you sure you want to delete cache for this workspace  and reload the window ?',
                 'Yes', 'Cancel');
             if (confirmation === 'Yes') {
-                    try{
-                        await stopClient(client);
-                        deactivated=true;
-                        await killNbProcess(false,log);
-                        await fs.promises.rmdir(userDir, {recursive : true});
-                        await vscode.window.showInformationMessage("Cache deleted successfully",'Reload window');
-                    }catch(err){
-                       await  vscode.window.showErrorMessage('Error deleting the cache','Reload window');
-                    }finally{
-                        vscode.commands.executeCommand("workbench.action.reloadWindow");
-                    }
+                try {
+                    await stopClient(client);
+                    deactivated = true;
+                    await killNbProcess(false, log);
+                    await fs.promises.rmdir(userDir, { recursive: true });
+                    await vscode.window.showInformationMessage("Cache deleted successfully", 'Reload window');
+                } catch (err) {
+                    await vscode.window.showErrorMessage('Error deleting the cache', 'Reload window');
+                } finally {
+                    vscode.commands.executeCommand("workbench.action.reloadWindow");
                 }
+            }
         } else {
             vscode.window.showErrorMessage('Cannot find userdir path');
         }
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -67,7 +67,7 @@ import { initializeRunConfiguration, runConfigurationProvider, runConfigurationN
 import { InputStep, MultiStepInput } from './utils';
 import { PropertiesView } from './propertiesView/propertiesView';
 import { openJDKSelectionView } from './jdkDownloader';
-
+import { NODE_WINDOWS_LABEL } from './constants';
 const API_VERSION : string = "1.0";
 const SERVER_NAME : string = "Oracle Java SE Language Server";
 export const COMMAND_PREFIX : string = "jdk";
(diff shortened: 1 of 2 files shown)


vscode/src/extension.ts
```diff
@@ -78,7 +78,7 @@ let nbProcess : ChildProcess | null = null;
 let debugPort: number = -1;
 let debugHash: string | undefined;
 let consoleLog: boolean = !!process.env['ENABLE_CONSOLE_LOG'];
-
+let deactivated = false;
```
Achal1607, line 81: Is this flag needed now?

vscode/src/extension.ts
```diff
@@ -514,16 +514,16 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
             if (confirmation === 'Yes') {
-                await fs.promises.rmdir(userDir, {recursive : true});
-                const res = await vscode.window.showInformationMessage('Cache cleared successfully for this workspace', 'Reload window');
-                if (res === 'Reload window') {
-                    await vscode.commands.executeCommand('workbench.action.reloadWindow');
-                }
+                stopClient(client).then(() => killNbProcess(false, log)).then(() => {
+                    fs.promises.rmdir(userDir, {recursive : true});
+                    vscode.commands.executeCommand('workbench.action.reloadWindow')
+                });
```
Achal1607, line 526: Maybe you can add a catch block here for better error handling.

vscode/src/extension.ts
```diff
@@ -514,16 +514,16 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
             vscode.window.showErrorMessage('Cannot find workspace path');
             return;
         }
+
```
Achal1607, line 517: We can maybe put this block of code in a try catch so that any error can be handled.

shivam71:
Added a callback for error handling

vscode/src/extension.ts
```diff
@@ -514,17 +514,19 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
         if (userDir && fs.existsSync(userDir)) {
-            const confirmation = await vscode.window.showInformationMessage('Are you sure you want to delete cache for this workspace?',
+            const confirmation = await vscode.window.showInformationMessage('Are you sure you want to delete cache for this workspace  and reload the window ?',
                 'Yes', 'Cancel');
             if (confirmation === 'Yes') {
-                await fs.promises.rmdir(userDir, {recursive : true});
-                const res = await vscode.window.showInformationMessage('Cache cleared successfully for this workspace', 'Reload window');
-                if (res === 'Reload window') {
-                    await vscode.commands.executeCommand('workbench.action.reloadWindow');
+                    stopClient(client).then(() => { deactivated=true;
```
Achal1607, line 523: Since the parent function is already using `async/await`, can we refactor this callback-based code to use `async/await` for better consistency? 
This will also allow us to handle errors more effectively using `try/catch` blocks.

vscode/src/extension.ts
```diff
@@ -953,7 +955,7 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
             if (p == nbProcess && code != 0 && code) {
                 vscode.window.showWarningMessage(`${SERVER_NAME} exited with ` + code);
             }
-            if (stdErr?.match(/Cannot find java/) || os.type() === "Windows_NT" ) {
+            if (stdErr?.match(/Cannot find java/) || (os.type() === "Windows_NT" && !deactivated) ) {
```
Achal1607, line 958: Can we move `"Windows_NT"` to constants file, since it is also used in #251 so that it can be utilized there as well?

Achal1607 (changes requested):

vscode/src/extension.ts
Achal1607: Try to add catch and finally

vscode/src/extension.ts
```diff
@@ -514,17 +514,24 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
-                await fs.promises.rmdir(userDir, {recursive : true});
-                const res = await vscode.window.showInformationMessage('Cache cleared successfully for this workspace', 'Reload window');
-                if (res === 'Reload window') {
-                    await vscode.commands.executeCommand('workbench.action.reloadWindow');
+                    try{
+                        await stopClient(client);
+                        deactivated=true;
+                        await killNbProcess(false,log);
+                        await fs.promises.rmdir(userDir, {recursive : true});
+                        await vscode.window.showInformationMessage("Cache deleted successfully",'Reload window');
```
Achal1607, line 528: nitpick: await not required here in my opinion

vscode/src/extension.ts
```diff
@@ -514,17 +514,24 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
+                    }catch(err){
+                       await  vscode.window.showErrorMessage('Error deleting the cache','Reload window');
```
Achal1607, line 530: nitpick: await not required here in my opinion

vscode/src/extension.ts
```diff
@@ -514,17 +514,24 @@ export function activate(context: ExtensionContext): VSNetBeansAPI {
...
         if (userDir && fs.existsSync(userDir)) {
-            const confirmation = await vscode.window.showInformationMessage('Are you sure you want to delete cache for this workspace?',
+            const confirmation = await vscode.window.showInformationMessage('Are you sure you want to delete cache for this workspace  and reload the window ?',
                 'Yes', 'Cancel');
             if (confirmation === 'Yes') {
-                await fs.promises.rmdir(userDir, {recursive : true});
-                const res = await vscode.window.showInformationMessage('Cache cleared successfully for this workspace', 'Reload window');
-                if (res === 'Reload window') {
-                    await vscode.commands.executeCommand('workbench.action.reloadWindow');
+                    try{
```
Achal1607, line 523: Can you please properly lint this block of code?

Achal1607 (changes requested):

vscode/src/extension.ts
shivam71: If I remove the await then the message may not be shown to the user "Cache deleted successfully" and the window will be reloaded . I tested without the await it doesn't work .

vscode/src/extension.ts
shivam71: Same as above

vscode/src/extension.ts
Achal1607: Sorry, my bad I thought it's just a simple information message, I noticed just now it is user action information message due to that await is required. Sorry, I should have seen in a more diligent way.

Achal1607 (approved):
LGTM. Thanks @shivam71 for making this change

PR #247, opened 2024-08-27
Backport NB23 patch 7390 to fix the Exception template

sid-srini:
In NB23, [patch #7390](https://github.com/apache/netbeans/pull/7390) fixed the regression where the newly created exception from template did not extend Exception, thereby failing to compile.

This temporarily backports the fix until NB23 is incorporated.


Achal1607 (approved):
Looks good to me. Thanks!

PR #248, opened 2024-08-28
Refactored JDK Downloader

Achal1607:
Improvements after refactoring JDK downloader:
- Transitioned from a functional to a class-based approach for better code organization and maintainability.
- Progress bar is also shown while downloading any JDK
- Improved error handling mechanisms
- Added comprehensive logging throughout the code to aid in debugging and increase traceability
- Replaced axios with node native library


THIRD_PARTY_LICENSES.txt
```diff
@@ -8656,217 +8656,6 @@ publicity pertaining to the work without specific, written prior permission.
 Title to copyright in this work will at all times remain with copyright 
 holders.
```
naren2605, line 8658: why s this getting removed?

vscode/src/jdkDownloader/resources/styles.ts
```diff
@@ -0,0 +1,103 @@
+export const downloaderCss = `
```
naren2605, line 1: we can probably follow naming convention like  <jdkDowloader>.styles.css , so that we can reuse same conventions

naren2605: idealy resources should not be part of code, it can be .properties .xml configs etc., see if we can we move this to .css file?

Achal1607: There are some limitations of vscode webview due to security purpose de don't allow us adding external references of resources, there is a work around for that, but I didn't wanted to use that that's why kept in .ts format

THIRD_PARTY_LICENSES.txt
Achal1607: axios is removed, so it's license is also removed from THIRD_PARTY license text file

vscode/src/jdkDownloader/resources/styles.ts
Achal1607: > we can probably follow naming convention like .styles.ts , so that we can reuse same conventions

Ok would rename the file. Thanks

vscode/src/jdkDownloader/downloader.ts
```diff
@@ -0,0 +1,263 @@
+import { commands, OpenDialogOptions, OutputChannel, ProgressLocation, window, workspace } from "vscode";
```
naren2605, line 1: can we rename it to downloader.action.ts -> as this would be absrtracting vcode actions

vscode/src/jdkDownloader/action.ts
```diff
@@ -0,0 +1,264 @@
...
+        const matchingOldDirs = await this.getMatchingDirs(this.jdkVersion!);
+        for await (const oldDirName of matchingOldDirs) {
+            await fs.promises.rmdir(path.join(this.DOWNLOAD_DIR, oldDirName), { recursive: true });
+        }
+    }
+
+    private extractJDK = async (): Promise<void> => {
+        this.logger.appendLine(`Extracting JDK...`);
+
+        const extractCommand = `tar -xzf "${this.downloadFilePath}" -C ${this.DOWNLOAD_DIR}`;
```
sid-srini, line 176: Do add quotes around the `DOWNLOAD_DIR` argument

sid-srini (changes requested):
A more reliable JDK download experience with progress notifications is a welcome addition. Thank you very much @Achal1607. And also for refactoring and cleaning up the code into view and control parts.

Couple of minor requests are mentioned inline.

Achal1607:
@sid-srini thank you for reviewing the PR. I have addressed the comments

sid-srini (approved):

vscode/l10n/bundle.l10n.en.json
```diff
@@ -16,13 +16,20 @@
...
   "jdk.downloader.label.selectOpenJdkVersion": "Select Oracle OpenJDK Version",
+  "jdk.downloader.message.downloadProgressBar":"{jdkType} {jdkVersion} download in progress",
   "jdk.downloader.message.downloadingAndCompletingSetup": "Downloading and completing setup of {jdkType} {jdkVersion}...",
   "jdk.downloader.message.downloadCompleted":"{jdkType} {jdkVersion} for {osType} download completed!",
   "jdk.downloader.error_message.whileSavingFile": "Error while saving the file: {error}",
   "jdk.downloader.error_message.whileDownloading": "Error while downloading the : {jdkType} {error}",
   "jdk.downloader.message.downloadFailed": "{jdkType} {jdkVersion} for {osType} download failed, wrong checksum.",
   "jdk.downloader.error_message.downloadFailedHttpError": "HTTP Error {statusCode} - {statusMessage}",
-  "jdk.downloader.error_message.jdkExtractionError": "Error: {error}",
+  "jdk.downloader.error_message.anyError": "Error: {error}",
```
naren2605, line 26: instead of anyError lets provide Some context

vscode/src/jdkDownloader/view.ts
```diff
@@ -0,0 +1,253 @@
+import { OPEN_JDK_VERSION_DOWNLOAD_LINKS, ORACLE_JDK_DOWNLOAD_VERSIONS } from '../constants';
+import { OutputChannel, Uri, ViewColumn, WebviewPanel, window } from 'vscode';
+import * as os from 'os';
+import { JdkDownloaderAction } from './action';
+import { downloaderCss } from './styles';
+import { l10n } from '../localiser';
+
+export class JdkDownloaderView {
+    public static readonly OPEN_JDK_LABEL = "OpenJDK";
```
naren2605, line 9: if these variables are not used outside we can mark these private

vscode/l10n/bundle.l10n.en.json
Achal1607: Actually I am not using this label anywhere, I have kept it since it was already present. So, should we remove this label or should I give a more generalized key like `jdk.extension.error_message.errorLabel`

vscode/src/jdkDownloader/action.ts
```diff
@@ -0,0 +1,299 @@
...
+    private machineArch?: string;
+    private installType?: string;
+    private installationPath?: string | null;
+    private downloadFilePath?: string;
+    private downloadUrl?: string;
+
+    constructor(private readonly logger: OutputChannel, private readonly downloaderView: JdkDownloaderView) { }
+
+    public attachListener = async (message: any) => {
+        const { command, id, jdkVersion, jdkOS, jdkArch, installType } = message;
```
naren2605, line 28: can we add parameters as  whatever we need in arguments? instead of entire message as it appears like we dont need entire message to be passed?

vscode/l10n/bundle.l10n.en.json
naren2605: ok lets leave it as is then

naren2605:
LGTM

naren2605 (approved):

PR #251, opened 2024-09-02
Restrict parent directory traversal for finding the associated project

sid-srini:
**Changes**
1. Added a launch-time JVM property for the netbeans LSP server "project.limitScanRoot", with its value set to the user's home directory path and the workspace folders, if any.
    - This is obtained using Node's `os.homedir()` function in *nbcode.ts*
    - Multiple paths are delimited by the OS-specific separator.
2. Added a configuration property "**jdk.advanced.disable.projectSearchLimit**" which can be used to turn off this new behaviour.

**Reasons**
- NetBeans supports searching for the build project information in the ancestor directories of the file or folder opened in a workspace.
- Thus, NetBeans `org.netbeans.modules.projectapi.SimpleFileOwnerQueryImplementation` supports a JVM launch property (i.e. System property) called **project.limitScanRoot**.
- This prevents searching for the build project information when the opened file or folder is outside this specified path.

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -68,7 +68,7 @@ import { InputStep, MultiStepInput } from './utils';
 import { PropertiesView } from './propertiesView/propertiesView';
 import { openJDKSelectionView } from './jdkDownloader';
 import { l10n } from './localiser';
-import { ORACLE_VSCODE_EXTENSION_ID,NODE_WINDOWS_LABEL } from './constants';
+import { ORACLE_VSCODE_EXTENSION_ID, NODE_WINDOWS_LABEL } from './constants';
 const API_VERSION : string = "1.0";
 const SERVER_NAME : string = "Oracle Java SE Language Server";
 export const COMMAND_PREFIX : string = "jdk";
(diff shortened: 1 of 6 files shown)


lahodaj:
When written like this, I see a problem if the user will have a folder in the workspace that is not under the user's home. For example, having a Maven project `/tmp/prj1` open as a workspace. The `(Simple)FileOwnerQuery` then will not be able to lookup the project, and things will work poorly. I think we should at the very least:
- permit search up until (any) workspace folder, if the folder is not inside the user's home,
- or: stop the search at user's home only if the search started inside the user's home,
- in addition to that: consider adding an option to disable this restriction in project search (although it may be difficult for the user's to understand that).

sid-srini:
> When written like this, I see a problem if the user will have a folder in the workspace that is not under the user's home. For example, having a Maven project `/tmp/prj1` open as a workspace. The `(Simple)FileOwnerQuery` then will not be able to lookup the project, and things will work poorly. I think we should at the very least:
> 
> * permit search up until (any) workspace folder, if the folder is not inside the user's home,
> * or: stop the search at user's home only if the search started inside the user's home,
> * in addition to that: consider adding an option to disable this restriction in project search (although it may be difficult for the user's to understand that).

Understood. 
- I will update for points 1 & 3 in this PR. 
- For point 2, I will be adding a separate PR in the NB repo and incorporate the patch once it is approved there.

vscode/src/extension.ts
```diff
@@ -902,11 +902,42 @@ function doActivateWithJDK(specifiedJDK: string | null, context: ExtensionContex
...
+                    }
+                    if (!projectFolderPath.startsWith(path.normalize(homeDir))) {
+                        userHomeDir = path.normalize(workspaces[0].uri.path);   // Keep userHomeDir limited to the workspace path.
+                    }
+                } catch (err:any) {
+                    handleLog(log, `Failed to check whether the workspace is inside the userHomeDir: ${err}`);
+                }
+            }
+        } else if (!userHomeDir) {
+            userHomeDir = os.type() === "Windows_NT" ? '%USERPROFILE%' : '$HOME';   // The launcher script may perform the env variable substitution
```
Achal1607, line 930: nitpick: Can we move `"Windows_NT"` to `constants.ts`, since it is used in #245 as well, so both can export from constants? (So one of the PRs may have to do this change, since I have made same comment on both the PRs).

sid-srini: Makes sense. Thanks. I'll rebase once #245 is merged.

Achal1607 (approved):
LGTM. Thanks @sid-srini

PR #266, opened 2024-09-12
Correctly pick configuration from launch settings during run and debug sessions

Achal1607:
This PR addresses an issue where the active launch configuration was not being correctly detected, leading to tests and other panels not receiving the proper settings during run and debug sessions. The changes ensure that the correct launch configuration is picked up, allowing for consistent and accurate behavior across all relevant panels.


sid-srini (approved):
Thanks @Achal1607. LGTM.

You may submit a separate PR for other possibly incorrect usages of invalid commandPrefix/extensionId in the extension.ts.

PR #273, opened 2024-09-16
[JAVSCODE-253] enabling nb support for localisation

naren2605:
enabling nb support for localisation https://github.com/oracle/javavscode/issues/253


build.xml
```diff
@@ -204,16 +264,30 @@
...
+            <property name="git-dir" value="netbeans"/>
+        </ant>
+        <ant target="git-patch">
+            <property name="git-patches-list" value="${patches-l10n}"/>
+            <property name="has-git-patches" value="${has-l10n-patches}"/>
+            <property name="git-dir" value="netbeans-l10n"/>
+        </ant>
+    </target>
+
     <target name="unapply-patches" if="has-patches">
```
sid-srini, line 291: Please add similar changes for `unapply-patches` as done for `apply-patches` above.

vscode/src/localiser.ts
```diff
@@ -29,6 +37,10 @@ class l10Wrapper implements l10n {
       const isPresentInBundle = valueFromBundle !== key;
       return isPresentInBundle ? valueFromBundle : this.defaultTranslation(key, placeholderMap);
     }
+    nbLocaleCode(){
+      const supportedLanguage = vcodeToNbLocaleCodes[vscode.env.language];
+      return supportedLanguage?supportedLanguage:DEFAULT_LANGAUGE;
```
sid-srini, line 42: It would be better for future support to replace this with a simple function. If required, at a later time, we could use such a mapping only for exceptional cases that do not fit in the standard transformation path.

Further, we do not need to manually set the language to `DEFAULT_LANGUAGE` since the NB backend may support other languages and we don't need to turn off this support. Anyhow, the NB backend will default to english if it does not find a message in the configured language.

So, I'd suggest making `vcodeToNbLocaleCodes = {}` for now, and, change this code block to something like below:
```suggestion
      const vscodeLanguage = vscode.env.language;
      if (!vscodeLanguage) return DEFAULT_LANGUAGE;
      let supportedLanguage:string = vcodeToNbLocaleCodes[vscodeLanguage];
      if (!supportedLanguage) {
          const localeParts = vscodeLanguage.split(/[-_]/g);
          if (localeParts.length > 1) {
              localeParts[1] = localeParts[1].toUpperCase();
          }
          supportedLanguage = localeParts.join(":");
      }
      return supportedLanguage;
```
Thanks.

sid-srini (commented):
Bringing netbeans localisation to the extension is a big enhancement. Thanks a lot @naren2605.

vscode/src/localiser.ts
naren2605: will remove vcodeToNbLocaleCodes for now, if requirement surfaces we can change it. otherwise it becomes funcitonally dead code.

sid-srini (approved):
Thanks. LGTM.

Achal1607 (commented):
Should we add update in `THIRD_PARTY_LICENSE.txt` as well?

Achal1607 (commented):
We are using currently following clusters from NetBeans. So maybe we can include only these cluster paths
```
    ${nbplatform.active.dir}/extide:
    ${nbplatform.active.dir}/harness:
    ${nbplatform.active.dir}/ide:
    ${nbplatform.active.dir}/java:
    ${nbplatform.active.dir}/nb:
    ${nbplatform.active.dir}/platform:
    ${nbplatform.active.dir}/webcommon
```

sid-srini (approved):

Achal1607 (approved):
LGTM. Thanks @naren2605

PR #281, opened 2024-09-19
Update jdk in github workflows config

sid-srini:
Changed jdk version from 17 to 21 since 17 is EOL now.


Achal1607 (approved):
LGTM. Thanks for noticing and updating this.

PR #289, opened 2024-10-03
Remove unneeded jar from netbeans modules

sid-srini:
Remove unneeded jar from netbeans modules
Corrected usage of disable.netbeans.l10n in build.xml


Achal1607 (approved):
LGTM. Thanks

PR #295, opened 2024-10-16
Adding netbeans and netbeans-l10n as submodules

shivam71:
Added netbeans and l10n as  submodules and  made required changes to support them.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "netbeans-l10n"]
+	path = netbeans-l10n
+	url = https://github.com/apache/netbeans-l10n.git
+[submodule "netbeans"]
+	path = netbeans
+	url = https://github.com/apache/netbeans.git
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "netbeans-l10n"]
+	path = netbeans-l10n
+	url = https://github.com/apache/netbeans-l10n.git
+[submodule "netbeans"]
```
naren2605, line 4: @shivam71  where can we mention specific commit id which checkout would be pointing to ?

netbeans
shivam71, line 1: @naren2605 Here the commit hash is stored

.gitmodules
shivam71: The .gitmodules just stored the url not the commit hash . See the changed files in this commit it shows netbeans also which is a special file just storing the commit hash.

naren2605:
LGTM

naren2605 (approved):

sid-srini (approved):
Thank you @shivam71 for improving the build experience and reducing human error in setup by getting this done. LGTM

PR #301, opened 2024-10-24
[JVSC-300] Remove quoting in nbcode launch arguments

### LINKED ISSUE #300: Oracle Java SE Language Server fails to initialize correctly in a directory containing spaces as of v23.0.0
As of v23.0.0, the language server fails to initialize correctly if a file is opened in a directory containing spaces. Modifying the directory names to avoid the use of spaces results in a successful initialization. However, reverting to v22.1.2 shows that the previous version of the extension could initialize the language server without problems, even for a file located in a directory containing spaces.

The error generated for a file in a directory with spaces, as of v23.0.0, is as follows

```
Initiating server activation
Request to kill LSP server.
Cannot kill: current process is None
Launching Oracle Java SE Language Server with /mnt/c/Users/invin/Desktop/Everything/Utilities/open-jdk-23.0.1 and userdir /home/Uinvin/.vscode-server/data/User/workspaceStorage/498fc74986182e8db43e45809883926d/Oracle.oracle-java
LSP server launching: 181
LSP server user directory: /home/Uinvin/.vscode-server/data/User/workspaceStorage/498fc74986182e8db43e45809883926d/Oracle.oracle-java
There are parameters but nobody wants to proces them: [2024/Data, Structures]
Cannot find org.netbeans.modules.java.lsp.server in the log!
Please refer to troubleshooting section for more info: https://github.com/oracle/javavscode/blob/main/README.md#troubleshooting
Request to kill LSP server.
Cannot kill: Requested kill on 181, current process is None
Oracle Java SE Language Server not enabled
```

Note that "2024" and "Data Structures" are part of the folder names that contain spaces.

sid-srini:
- The nbcode executable is launched from the ts/js extension using Node's `child_process.spawn` function.
    - This function directly launches the command in a new process. It does NOT create a shell to execute it; unlike `child_process.exec`.
- Therefore, quoting the arguments passed to prevent shell splitting is not required.
    - In fact, due to the presence of extra quotes, it may lead to incorrect splitting of arguments with spaces in them.
    - This is the cause for the initialization failure of the extension nbls when the project folder opened contains spaces in its path.
- Fixed *nbcode.ts* to remove the extra quotes added for some of the option values.

Closes #300

### DIFF
--- a/vscode/src/nbcode.ts
+++ b/vscode/src/nbcode.ts
@@ -105,13 +105,13 @@ export function launch(
     }
 
     if (info.projectSearchRoots) {
-        ideArgs.push(`-J-Dproject.limitScanRoot="${info.projectSearchRoots}"`);
+        ideArgs.push(`-J-Dproject.limitScanRoot=${info.projectSearchRoots}`);
     }
 
     if (info.verbose) {
         ideArgs.push('-J-Dnetbeans.logger.console=true');
     }
-    ideArgs.push(`-J-Dnetbeans.extra.dirs="${clusterPath}"`)
+    ideArgs.push(`-J-Dnetbeans.extra.dirs=${clusterPath}`)
     if (env['netbeans.extra.options']) {     
         ideArgs.push(...env['netbeans.extra.options'].split(' '));
     }


Achal1607 (approved):
LGTM. Thanks

PR #303, opened 2024-10-26
Fix Netbeans tests failures

sid-srini:
1. Adding the patch for netbeans PR #7910 to prevent tests failures due to CRLF line endings.
2. Added the use of the extension configuration prefix `"jdk"` in *ServerTest*, as had been done for *java.lsp.server* *AttachConfigurations* and *ProjectConfigurationCompletion*, via the *patch/rename-debugger.diff*.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -55,6 +55,7 @@
             patches/7722.diff
             patches/7724.diff
             patches/7733.diff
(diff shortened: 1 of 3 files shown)


Achal1607 (approved):
LGTM. Thanks!

Issue #217, opened 2024-07-26
test runner incorrectly processes test names


robaho:
If you have two tests named with the same prefix, the test runner shows them as nested.

For example, `ensureAllElementsAreRead` and `ensureAllElementsAreReadUsingThread` it shows the latter nested under the first.

See attached screenshot.
<img width="1259" alt="Screen Shot 2024-07-25 at 9 42 45 AM" src="https://github.com/user-attachments/assets/e4617218-dc32-49f7-8d86-8ab1c37f6e9b">

sid-srini:
Hi @robaho. Thanks for providing a screenshot of the issue. Would you please provide some more details about your environment? such as:
1. The version of VS Code you're using;
2. The version of the extension you're using;
3. The version of JUnit you are using;
4. Whether the project is a maven or a gradle project; and which version is in use.
5. Any other information that you may think might be useful for us to reproduce this issue. 

Thank you.

robaho:
```
Version: 1.91.1 (Universal)
Commit: f1e16e1e6214d7c44d078b1f0607b2388f29d729
Date: 2024-07-09T22:07:54.982Z
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin x64 21.6.0
```

Oracle extension version v22.1.1
Gradle
JDK 21

As I mentioned in the related issue, you can install the project from https://github.com/robaho/closablequeue

naren2605:
rca: issue is due to grouping based on prefix and leaving extra entries in explorer panel which are not getting right test status updates
 
 fix:  will be removing grouping

PR #307, opened 2024-10-29
JVSC #217 | fixing test runner incorrectly processes test names

naren2605:
fix related details mentioned here https://github.com/oracle/javavscode/issues/217#issuecomment-2443458307 

sample test run after fixing the issue:

image

### DIFF
--- a/vscode/src/views/TestViewController.ts
+++ b/vscode/src/views/TestViewController.ts
@@ -25,6 +25,7 @@ import { commands, debug, tests, workspace, CancellationToken, TestController, T
 import * as path from 'path';
 import { asRange, TestCase, TestSuite } from "../lsp/protocol";
 import { extCommands, builtInCommands, nbCommands } from "../commands/commands"
+import { extConstants } from "../constants";
 
 export class NbTestAdapter {
 
@@ -35,7 +36,7 @@ export class NbTestAdapter {
     private started: boolean = false;
 
     constructor() {
-        this.testController = tests.createTestController('apacheNetBeansController', 'Apache NetBeans');
+        this.testController = tests.createTestController(extConstants.ORACLE_VSCODE_EXTENSION_ID + '.testController', 'Java');
         const runHandler = (request: TestRunRequest, cancellation: CancellationToken) => this.run(request, cancellation);
         this.testController.createRunProfile('Run Tests', TestRunProfileKind.Run, runHandler);
         this.testController.createRunProfile('Debug Tests', TestRunProfileKind.Debug, runHandler);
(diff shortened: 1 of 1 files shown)


vscode/src/views/TestViewController.ts
```diff
@@ -23,7 +23,7 @@
 
 import { commands, debug, tests, workspace, CancellationToken, TestController, TestItem, TestRunProfileKind, TestRunRequest, Uri, TestRun, TestMessage, Location, Position, MarkdownString } from "vscode";
 import * as path from 'path';
-import { asRange, TestCase, TestSuite } from "../lsp/protocol";
+import { asRange, TestCase as NBTestCase, TestSuite as NbTestSuite } from "../lsp/protocol";
```
sid-srini, line 26: Is there any particular reason for this name change?
- Logically, I don't think NB should be prefixed, since its not a Netbeans test-case/suite, but interfaces to the LSP protocol.
- The camel-casing becomes tedious and is inconsistent between `NB` and `Nb`.
- A lot of code changes in this PR are just for the renaming and not the fixes, adding a kind of "noise".

As per my understanding, only the changes on lines 38, 238-240 and 246, seem necessary.

vscode/src/views/TestViewController.ts
```diff
@@ -35,7 +35,7 @@ export class NbTestAdapter {
     private started: boolean = false;
 
     constructor() {
-        this.testController = tests.createTestController('apacheNetBeansController', 'Apache NetBeans');
+        this.testController = tests.createTestController('apacheNetBeansController', 'Oracle Java VS code');
```
sid-srini, line 38: 1. The api doc for [createTestController](https://code.visualstudio.com/api/references/vscode-api#tests) mentions that the `id` (i.e. first arg) should be globally unique.
    - Perhaps we should use: `extConstants.ORACLE_VSCODE_EXTENSION_ID + '.testController'`
2. The user-visible label needs to be in line with our extension name , i.e. `Java`.

vscode/src/views/TestViewController.ts
```diff
@@ -109,7 +109,9 @@ export class NbTestAdapter {
                     break;
             }
             if (!noPassDown) {
-                item.children.forEach(child => this.set(child, state, message, noPassDown));
+                item.children.forEach(child =>
+                     this.set(child, state, message, noPassDown)
+                    );
```
sid-srini, line 114: reformatting seems unnecessary here.

vscode/src/views/TestViewController.ts
```diff
@@ -199,79 +201,54 @@ export class NbTestAdapter {
...
-                        if (!arr) {
-                            parentTests.set(parent.test, arr = []);
-                            children.push(parent.test);
-                        }
-                        arr.push(this.testController.createTestItem(test.id, parent.label));
-                    }
-                } else {
-                    currentTest = this.testController.createTestItem(test.id, test.name, testUri);
-                    currentTest.range = asRange(test.range);
+                if(!testExecution){
```
sid-srini, line 238: nitpick:
```suggestion
                if (!testExecution) {
```

vscode/src/views/TestViewController.ts
```diff
@@ -199,79 +201,54 @@ export class NbTestAdapter {
...
         });
-        if (testExecution) {
-            parentTests.forEach((val, key) => {
-                const item = this.testController.createTestItem(key.id, key.label, key.uri);
-                item.range = key.range;
-                item.children.replace(val);
-                currentSuite?.children.add(item);
-            });
-        } else {
+        if(!testExecution)  {
```
sid-srini, line 246: nitpick:
```suggestion
        if (!testExecution) {
```

sid-srini (changes requested):
Thanks @naren2605 for fixing this so quickly. 

Please check my comments on reducing the scope of the PR changes to the required changes only. Thanks.

Achal1607 (approved):
LGTM. Thanks!

sid-srini (approved):
This is very useful, especially for your added fix on the test controller creation. 👍 Thank you very much @naren2605.

PR #309, opened 2024-10-29
Patch for  MoveRefactoring null error fix

shivam71:
- The MoveRefactoring with the default visibility option as "Escalate" was not working  . 
- The issue seems to be from an uncaught null pointer exception arising somewhere else . 
- It got fixed by minimal changes on NB . 
Netbeans PR link [#7923](https://github.com/apache/netbeans/pull/7923)


sid-srini (commented):
Thanks. Looks fine to me. 👍 

Would you be able to confirm if there are any tests in the netbeans test cases that could cover this NPE flow? Or whether a small test case could be added as part of the upstream PR?

Achal1607 (commented):
Can we add any tests around this on NetBeans side?

shivam71:
Hmm I will check for the tests though the change seems almost harmless we are catching an exception which was not being caught were the later is more dangerous . I am not sure of the tests but this NPE is always caught because initially an empty HTML form gets created with none of the fields selected so this always happens .

build.xml
```diff
@@ -58,6 +58,7 @@
             patches/7750.diff
             patches/7910.diff
             patches/7921.diff
+            patches/7923.diff
```
sid-srini, line 61: Would you please rename this file to something like *7923-draft.diff* so that we can remember to update the patch when the upstream PR is approved? Thanks.

sid-srini (changes requested):

build.xml
shivam71: Sure !

sid-srini (approved):
Thanks @shivam71. Looks ok to me for the interim, until the upstream PR is reviewed and approved. At that point, please update the patch with any changes, as a fresh PR here.

PR #313, opened 2024-10-30
Fix Netbeans tests failures on MacOS

sid-srini:
Adding the patch for the netbeans PR apache/netbeans#7926, to fix the unit test failures in the java.hints module due to old Apple JDK 6 special treatment for the expected output.


Achal1607 (dismissed):
LGTM. Thanks!

Achal1607 (approved):
LGTM. Thanks!

PR #314, opened 2024-10-30
Fix NetBeans launch failure on unix shell for arguments with quotes et al

sid-srini:
Backporting Netbeans PR #7908 (still under review) in order for the LSP server launch to work for launch arguments including the project path containing special characters like quotes, spaces, backticks and $.

This is an interim patch to mitigate the effect on user experience, while the upstream PR is under review. It will be updated to the approved patch, as and when it is made available.

This is partially related to #301, which only fixes for spaces in arguments.

Additionally, the same diff is applied to script/bin/nbcode which is a copy of the netbeans java.lsp.server module's copy.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -56,6 +56,7 @@
             patches/7724.diff
             patches/7733.diff
             patches/7750.diff
+            patches/7908-draft.diff
             patches/7910.diff
             patches/7921.diff
             patches/7923_draft.diff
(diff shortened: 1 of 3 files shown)


Achal1607 (approved):
LGTM. I am not very familiar with shell scripts so please @naren2605 if you can have a look then it would be great.

Asad-Mirza-24 (commented):
hmmm

donphelix (approved):
Looks okay to me, just resolve the conflicts and it's good to go

singh-akhilesh (approved):
Looks reasonable to me.

sid-srini:
Completed with changes as part of v26.0.1; closing this PR.

PR #318, opened 2024-11-01
Disable SecurityManager

lahodaj:
As consequence of JEP-486, it will not be possible to set SecurityManager. The JEP is currently Proposed to Target for JDK 24:
https://mail.openjdk.org/pipermail/jdk-dev/2024-November/009601.html

The NetBeans back we use is currently setting a SecurityManager, and will fail to start with the updated JDK. Inside the NetBeans project, there's a motion to remove the SecurityManager:
https://github.com/apache/netbeans/pull/7928

This patch proposed herein is a temporary attempt to disable the SecurityManager inside the extension, until the above PR is finished in one way or another, and until we upgrade to the version of NetBeans that will include it.


sid-srini (approved):
Thanks a lot @lahodaj for incorporating this. LGTM 👍

PR #322, opened 2024-11-06
Upgraded dev dependencies and removed unused dependencies

Achal1607:
### Upgraded dev dependencies
- @types/node: 13.11.0 -> 18.19.64
- typescript: 4.6.4 -> 5.3.3
- @types/mocha: 9.0.0 -> 10.0.9
- mocha: 9.1.2 -> 10.8.2
- @types/vscode: 1.76.0 -> 1.84.0
- @vscode/codicons: 0.0.29 -> 0.0.36
- esbuild: 0.16.17 -> 0.24.0

### Upgraded VS Code engine
- 1.76.0 -> 1.84.0

### Removed unused dev dependencies
- ps-node
- @types/vscode-webview => Provides types for webviews , but currently not used anywhere in the code. Types provided: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vscode-webview/index.d.ts

### Replaced `fs.rmdir()` to `fs.rm()`
Deprecation warning of `rmdir()` was showing up earlier before upgrade of @types/node. 
```
DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vscode-webview/index.d.ts 

So, replaced instances of `rmdir` with `rm`.

### Updated package-lock.json
Deleted package-lock.json and created again (that's why it's version is changed from 2 to 3) so that it removes all the earlier redundant entries and generates a new file on the basis of updated package.json.

### DIFF
--- a/patches/dev-dependency-licenses.diff
+++ b/patches/dev-dependency-licenses.diff
@@ -136,3 +136,30 @@ index 0000000000..803b55c862
 +LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 +OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 +THE SOFTWARE.
+diff --git a/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license b/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license
+new file mode 100644
+index 0000000000..e7323bb52e
+--- /dev/null
++++ b/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license
+@@ -0,0 +1,21 @@
++MIT License
++
++Copyright (c) Matteo Collina and Undici contributors
++
++Permission is hereby granted, free of charge, to any person obtaining a copy
++of this software and associated documentation files (the "Software"), to deal
++in the Software without restriction, including without limitation the rights
++to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
++copies of the Software, and to permit persons to whom the Software is
++furnished to do so, subject to the following conditions:
++
++The above copyright notice and this permission notice shall be included in all
++copies or substantial portions of the Software.
++
++THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
++IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
++FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


vscode/package.json
```diff
@@ -776,24 +776,21 @@
 		"@istanbuljs/nyc-config-typescript": "^1.0.2",
 		"@types/chai": "^4.3.17",
 		"@types/glob": "^7.1.1",
-		"@types/mocha": "^9.0.0",
-		"@types/node": "^13.11.0",
-		"@types/ps-node": "^0.1.0",
```
naren2605, line 781: this is removed intentionally right?

patches/dev-dependency-licenses.diff
```diff
@@ -136,3 +136,30 @@ index 0000000000..803b55c862
 +LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 +OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 +THE SOFTWARE.
+diff --git a/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license b/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license
+new file mode 100644
+index 0000000000..e7323bb52e
+--- /dev/null
++++ b/nbbuild/misc/prepare-bundles/src/main/resources/org/netbeans/prepare/bundles/undici-types-5.26.5-license
```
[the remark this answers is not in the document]
Achal1607: Since, typescript was updated it's dependency also got upgraded. License for that is included in the patch.

vscode/package-lock.json
```diff
@@ -415,58 +412,10 @@
...
-				"node": ">=12"
-			}
-		},
 		"node_modules/@esbuild/darwin-arm64": {
-			"version": "0.16.17",
-			"resolved": "https://registry.npmjs.org/@esbuild/darwin-arm64/-/darwin-arm64-0.16.17.tgz",
-			"integrity": "sha512-/2agbUEfmxWHi9ARTX6OQ/KgXnOWfsNlTeLcoV7HSuSTv63E4DqtAc+2XqGw1KHxKMHGZgbVCZge7HXWX9Vn+w==",
+			"version": "0.24.0",
+			"resolved": "https://registry.npmjs.org/@esbuild/darwin-arm64/-/darwin-arm64-0.24.0.tgz",
+			"integrity": "sha512-CKyDpRbK1hXwv79soeTJNHb5EiG6ct3efd/FTPdzOWdbZZfGhpbcqIpiD0+vwmpu0wTIL97ZRPZu8vUt46nBSw==",
```
sid-srini, line 418: Would it make sense to remove this also from the package-lock? 
The `esbuild` run on the relevant os-arch host would get the required dependency, right?

naren2605 (approved):

vscode/package-lock.json
Achal1607: I don't know but even if we can in my opinion we shouldn't update anything manually in package-lock.json since it is auto generated,

PR #292, opened 2024-10-11
Refactored codebase

Achal1607:
The main aim of refactoring was to make code more modular so that it is more readable, scalable and maintainable. 
- extension.ts is refactored into multiple modules
- LSP server startup is simplified
- commands and configuration modules are created
- Debugger is separated out in a different module
- Logger and other utils files are created.

NOTE: Views are yet to be refactored which can be done in a separate PR or included in this only once it is done.

PR #328, opened 2024-11-12
Fixed broken debugger attach UI options

Achal1607:
Attach to port, Attach to process, etc. options available in the Debugger Session seemed to be broken after refactoring the codebase (https://github.com/oracle/javavscode/pull/292).
So, this PR aims to fix the broken configuration.


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #339, opened 2024-12-09
Changes to maintain a single README file.

shivam71:
- Removed the README inside the vscode folder
- Added ant tasks for creating the README for packaging 
-  Excluded images folder from the vsix as it was not being used 

To test the README that will be created and uploaded on marketplace follow the steps :-
1. Build the vscode ext with `ant build-vscode-ext`.
2. Unzip the vsix .(change extension to `.zip` and then unzip.)
3. Open the README in the vsix and preview the markdown as well as check the relative urls in the markdown itself.


Achal1607 (approved):
LGTM. Thanks!

build.xml
```diff
@@ -246,19 +246,36 @@
         
         <copy file="${basedir}/LICENSE.txt" todir="${basedir}/vscode" />
         <copy file="${basedir}/THIRD_PARTY_LICENSES.txt" todir="${basedir}/vscode" />
-        
+        <copy file="${basedir}/README.md" todir="${basedir}/vscode" />
```
naren2605, line 249: @shivam71 i suppose image resource paths also needs to be corrected accordingly when we move readme.md from root dir to vscode dir. could you check once?
`Enable Preview has to  get converted to Enable Preview `

naren2605: or bit cleaner approach is  we can maintain images in root dir level as a seperate dir and copy images dir also to vscode dir - this might need bit of refactoring.

shivam71: Thanks for noticing that ! , since while packaging the relative urls are converted to https GitHub urls this isn't an issue .

Dexxxto (approved):

PR #346, opened 2024-12-11
Minor Code Fixes

subhash-arabhi:
Patches from following PRs
[#8038](https://github.com/apache/netbeans/pull/8038)
[#8036](https://github.com/apache/netbeans/pull/8036)


build.xml
```diff
@@ -60,6 +60,8 @@
             patches/7921.diff
             patches/7923.diff
             patches/7926.diff
+            patches/8036.diff
+            patches/8038.diff
```
sid-srini, line 64: Please add a `-draft` suffix to the patch filenames so that it reminds us to track and update the final approved version as and when it happens.

sid-srini (changes requested):
Thanks @subhash-arabhi.

PR #349, opened 2024-12-13
Updated the labels associated with disable/enable NbJavac shown in the settings and notifications.

shivam71:
- With Netbeans 23 the extension requires javac from JDK23+ or fallback to nb-javac if not available.
- Hence the labels associated with enable/disable Nbjavac need to state the minimum required JDK to be 23.
- Labels present in l10n bundles and package.nls updated across the supported languages .


sid-srini (approved):
Thanks @shivam71. LGTM 👍

PR #355, opened 2024-12-26
Updates on 8036 patch

subhash-arabhi:
Updated the [#8036](https://github.com/apache/netbeans/pull/8036) based on review comments


sid-srini (approved):
Thanks @subhash-arabhi. LGTM 👍

PR #357, opened 2024-12-27
Increased integration tests timeouts

Achal1607:
Sometimes tests fail due to timeout issues so increasing the timeouts of integration tests for better maintenance of the codebase.


sid-srini (approved):
Thanks @Achal1607. LGTM 👍 
In a future PR, you may try to make this a configurable value which may be set via the npm command-line. Thanks.

PR #350, opened 2024-12-14
Adding initial structure of telemetry jdkFeatures event generation

sid-srini:
Addition of the java.lsp.server.telemetry module to obtain Java language features from the NetBeans LSP Server.

PR #360, opened 2024-12-28
Fixes and edits to telemetry code to support JdkFeatureEvent

sid-srini:
Prepare the existing telemetry framework to support the handling and reporting of jdkFeature events. 
Also, amended utilities to aid in populating fields of the jdkFeature event.

Note: this is the preliminary commit in the PR #350 which provides an implementation for the generation of the jdkFeature event.

### DIFF
--- a/vscode/src/lsp/listeners/notifications/handlers.ts
+++ b/vscode/src/lsp/listeners/notifications/handlers.ts
@@ -25,6 +25,7 @@ import { LOGGER } from '../../../logger';
 import { globalState } from "../../../globalState";
 import { WorkspaceChangeData, WorkspaceChangeEvent } from "../../../telemetry/events/workspaceChange";
 import { Telemetry } from "../../../telemetry/telemetry";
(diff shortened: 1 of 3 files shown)


vscode/src/lsp/listeners/notifications/handlers.ts
```diff
@@ -123,6 +124,17 @@ const telemetryEventHandler = (param: any) => {
         Telemetry.sendTelemetry(workspaceChangeEvent);
         return;
     }
+    if (JdkFeatureEvent.NAME === param?.name) {
+        const {javaVersion, names, jeps, isPreviewEnabled} = param?.properties;
```
Achal1607, line 128: nipick: Can you please format this block of code?

Achal1607 (changes requested):
There is only one minor change. Otherwise it looks fine. Thanks!

Achal1607 (approved):
LGTM. Thanks

PR #364, opened 2025-01-13
Fixes to TelemetryServerManager to ensure multi-module projects under a workspace root are recognized

sid-srini:
1. Amended `LspServerTelemetryManager.sendWorkspaceInfo` to use a `NavigableMap` to find sub-paths of the workspaceFolder that correspond to Projects, when it is not itself a Project.
2. Also, caught exceptions from `ProjectProblems.isBroken()` since it maybe a sign of a broken project.


Achal1607 (approved):
LGTM. Thanks for fixing it!

PR #370, opened 2025-02-11
Updated vsix building in order to pass vsixsigntool validation after modified codesigning flow

sid-srini:
The `vsixsigntool` has certain pre-requisites for its verification to pass. These are no longer done in the code-signing flow, requiring them to be performed prior to the creation of the vsix. 

It seems that the VS Code Marketplace publishing uses a different validation flow for signed vsix, since the unsigned vsix created by the [@vscode/vsce](@vscode/vsce) node tool doesn't adhere to these restrictions.
For example:
- `[Content_Types].xml` must contain file-extension strings without a `'.'` prefix in a signed vsix-extension.
- There must be no files present in the vsix-extension without a file-extension.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -63,6 +63,7 @@
             patches/8036-draft.diff
             patches/8038-draft.diff
             patches/mvn-sh.diff
+            patches/project-marker-jdk.diff
             patches/generate-dependencies.diff
             patches/rename-debugger.diff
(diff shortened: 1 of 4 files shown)


lahodaj (approved):

Achal1607 (approved):
LGTM. Thanks

PR #376, opened 2025-02-17
Automate dropdown options of OracleJDK version in JDK downloader

Achal1607:
Automate options available in OracleJDK download options in JDK Downloader.
Also, removed a test which was redundant. It was directly testing the stub instead of the original method due to which race condition is happening and it is failing.

### DIFF
--- a/vscode/src/constants.ts
+++ b/vscode/src/constants.ts
@@ -26,12 +26,13 @@ export namespace extConstants {
 }
 
 export namespace jdkDownloaderConstants {
-  export const JDK_RELEASES_TRACK_URL = `https://www.java.com/releases/releases.json`;
+  
+  export const ORACLE_JDK_RELEASES_BASE_URL = `https://java.oraclecloud.com/currentJavaReleases`;
 
   export const ORACLE_JDK_BASE_DOWNLOAD_URL = `https://download.oracle.com/java`;
 
-  export const ORACLE_JDK_DOWNLOAD_VERSIONS = ['23', '21'];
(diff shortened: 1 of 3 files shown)


sid-srini (changes requested):
Thanks @Achal1607 . This will significantly improve jdk access. One small suggestion would be to maintain the old jdk versions list as a fallback, in case the remote fetch fails due to some external cause or API change.

arvindaprameya:
[like]  Arvind Aprameya reacted to your message:
________________________________
From: Siddharth Srinivasan ***@***.***>
Sent: Tuesday, February 18, 2025 4:50:45 AM
To: oracle/javavscode ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [External] : Re: [oracle/javavscode] Automate dropdown options of OracleJDK version in JDK downloader (PR #376)


@sid-srini requested changes on this pull request.

Thanks @Achal1607<https://urldefense.com/v3/__https://github.com/Achal1607__;!!ACWV5N9M2RV99hQ!K1foODTWP5-5p3r7VwlL4X6wDuSOXtcjVjJiru0FgH_6rVKupA6ZaLRbn3kPeSkma3dIfl2prgEfrN4RtPRsKgkx-DGg$> . This will significantly improve jdk access. One small suggestion would be to maintain the old jdk versions list as a fallback, in case the remote fetch fails due to some external cause or API change.

sid-srini (approved):
Thanks. LGTM :+1

PR #379, opened 2025-02-25
Fixed label and id issue in JDK downloader

Achal1607:
- Label constant was being used as id in JDK downloader which can be a issue if labels are localised. 
- So, separated out labels and Ids so that localisation doesn't affect the Id creation.


sid-srini (approved):
Thanks @Achal1607. LGTM :+1:

PR #381, opened 2025-03-04
Fix build script which gets stuck sometimes

Achal1607:
Currently there's an issue while applying patch to vsce, the build gets stuck. This PR aims to solve this issue by introducing new target which first checks the patch and then tries to apply if not already applied.

Thanks @sid-srini for helping fix this issue.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -208,8 +208,38 @@
         </ant>
     </target>
 
-
-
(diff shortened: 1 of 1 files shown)


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #382, opened 2025-03-04
Update to NB-25

Achal1607:
This PR aims to update LS to NB-25.


vscode/src/lsp/protocol.ts
sid-srini, line 1: Are these commands `'output/*'` being triggered from the UI? Do they need to be added to *package.json*?

sid-srini (commented):
Thanks a lot @Achal1607 for upgrading the LS to NB25.

patches/6330.diff
```diff
@@ -146,15 +144,15 @@ index 9ab389564c..9b7c044c81 100644
...
              ByteArrayInputStream in = new ByteArrayInputStream(bytes);
              ByteArrayOutputStream os = new ByteArrayOutputStream();
              conn.prepare("Pipe server", in, os, new LspSession(), ConnectionSpecTest::setCopy, ConnectionSpecTest::copy);
 -            String reply = os.toString("UTF-8");
-+            String reply = os.toString("UTF-8").replaceAll("\n$", "");
++            String reply = os.toString("UTF-8".replaceAll("\n$", ""));
              String exp = "Pipe server listening at port ";
              assertTrue(reply, reply.startsWith(exp));
-             int port = Integer.parseInt(reply.substring(exp.length()));
+             int port = Integer.parseInt(reply.substring(exp.length(), reply.indexOf('\n', exp.length())));
```
sid-srini, line 158: This also appears to be a buggy change, which is better than the original, but still not a complete fix.
```suggestion
             int nextLineIdx = reply.indexOf('\n', exp.length());
             if (nextLineIdx < 0) nextLineIdx = reply.length();
             int port = Integer.parseInt(reply.substring(exp.length(), nextLineIdx));
```

vscode/src/lsp/protocol.ts
Achal1607: No, it is used for the identification of the backend only like what protocol the frontend is sending, so it doesn't need to be added to package.json

vscode/package.json
```diff
@@ -290,8 +295,13 @@
 							"vmArgs": {
 								"type": [
 									"string",
+									"array",
 									"null"
 								],
+								"items": {
+									"type": "string",
+									"description": "%jdk.debugger.configuration.vmArgs.items.description%"
```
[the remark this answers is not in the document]
Achal1607: In our current README, we haven't explicitly mentioned anywhere about `vmArgs` type i.e. `vmArgs` is a `string` type, we haven't mentioned.

patches/6330.diff
Achal1607: This was implemented in the following NetBeans commit. However, your approach is better, so I have incorporated it as well. 
https://github.com/apache/netbeans/commit/29b40aae13ef27c7073dc3b2609b8f1b9ebb3877

vscode/src/debugger/debugger.ts
```diff
@@ -237,7 +237,11 @@ class RunConfigurationProvider implements vscode.DebugConfigurationProvider {
             if (vmArgs) {
                 if (!config.vmArgs) {
                     config.vmArgs = vmArgs;
+                } else if (Array.isArray(config.vmArgs)) {
+                    let cfg: string[] = config.vmArgs;
+                    cfg.push(vmArgs);
```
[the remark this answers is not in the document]
Achal1607: `vmArgs` is always a string since it comes from the `settings.json`. In package.json you can find:
```
"jdk.runConfig.vmOptions": {
	"type": "string",
	"default": "",
	"description": "%jdk.configuration.vmOptions.description%"
}
```
So, we don't need to do any string manipulation there IMO.
launch.json can have `array` of `vmArgs`. For example:
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "jdk",
            "request": "launch",
            "name": "Launch Java App",
            "vmArgs": ["-Xms256m"]
        }
    ]
}
```
But the side panel can only have `string` type of `vmOptions` as it comes from the `settings.json`.  
image

In future can also convert `settings.json` configuration to take input as array as well.

Achal1607:
@sid-srini I have addressed all the review comments. Please review it again. Thanks

vscode/src/debugger/debugger.ts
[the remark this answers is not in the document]
sid-srini: Ok. Then we would need to do just the following:
> In case vmArgs is a string and is to be pushed into an array, then we should perhaps split using quote-aware whitespace-delimiters into an array before pushing. That would ensure that downstream usage of config.vmArgs does not cause the double-quoting of multiple args into a single arg, which would not be understood by the JVM.

vscode/src/lsp/protocol.ts
sid-srini: Thanks for clarifying.

vscode/src/utils.ts
```diff
@@ -301,4 +301,39 @@ export async function initializeRunConfiguration(): Promise<boolean> {
...
+			result.push(current);
+			current = "";
+		} else if (isQuotes(char)) {
+			const quoteType = char;
+			current += char;
+			i++;
+			let f = true;
+			while (i < input.length && f) {
+				current += input[i];
+				if (input[i] === quoteType && input[i - 1] != "\\")
```
sid-srini, line 325: Typically, this will be functionally correct, although the performance maybe improved by storing a regex reference and performing a `search` followed by `concat` of `substr`. However, in certain corner cases this may not give the correct result. For e.g. when there are a balanced number of escape backslashes which indicate a literal backslash instead: 
- `"abc \\" def` should become 2 strings; whereas the current code would retain it as a single string.

vscode/src/utils.ts
```diff
@@ -301,4 +301,39 @@ export async function initializeRunConfiguration(): Promise<boolean> {
...
-}
\ No newline at end of file
+}
+
+const isQuotes = (c: string): boolean => {
+	return c === "'" || c === '"';
+}
+
+export const parseArguments = (input: string): string[] => {
+	const result: string[] = [];
```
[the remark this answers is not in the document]
Achal1607: It is a string type field and if someone is typing as an Array then IMO it's the user issue and we shouldn't check that.

vscode/src/utils.ts
Achal1607: Thanks for pointing out this edge case, will update the method.

vscode/src/utils.ts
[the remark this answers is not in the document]
sid-srini: No no, this is not to do with the value being a string or not. The UI forces the user to input multiple options in a single string with space-separators. 
- In the typical use-case, each option will not contain any spaces within it. Thus, we can optimise the flow by quickly checking for the absence of quotes and split on spaces.
- The more complex (and also performance intensive) code would then run only in the atypical scenarios when users have provided space-containing options by utilising quoting.

Achal1607:
@sid-srini Addressed review comments. Thanks

vscode/src/utils.ts
```diff
@@ -301,4 +301,42 @@ export async function initializeRunConfiguration(): Promise<boolean> {
...
+			result.push(current);
+			current = "";
+		} else if (isQuotes(char)) {
+			const quoteType = char;
+			current += char;
+			i++;
+			let f = true;
+			while (i < input.length && f) {
+				current += input[i];
+				const isEscapingSomethingElse = (i > 1 && input[i - 1] == "\\" && input[i - 2] == "\\");
```
sid-srini, line 326: Nit: an index out-of-bounds error may occur with the unchecked `i-1` or `i-2` index access. Safety or performance-wise, it may be better to add a separate else-if block for keeping track of number of continuous occurrences of `char == '\\'` i.e. `isEscapingActive` and just check that during quote handling.

sid-srini (approved):
Thanks @Achal1607 for completing the upgrade to NB25 and also improving the implementation of some of the new functionality. LGTM 👍

vscode/src/utils.ts
Achal1607: `i > 1` will check if index is greater than 1 if it is true then only javascript will check other conditions. 

image

Please correct me if I perceived your comment incorrectly?

PR #384, opened 2025-03-07
Changed the patch check argument in the build file

sid-srini:
Changed the patch check argument in the build file from `--check` to `--dry-run` for wider os compatibility when used for patching vsce for vsixsign compatibility.


Achal1607 (approved):
LGTM. Thank you @sid-srini

PR #401, opened 2025-04-16
Updated test timeouts and improved maintainability of them

Achal1607:
Removed individual tests timeouts and shifted to a single place so that it is easy to maintain and also changed from `milliseconds` to `string` for better readability


sid-srini (approved):
LGTM 👍 . Thanks

PR #403, opened 2025-04-18
Enhancement in telemetry preferences flow and some telemetry maintenance changes 

Achal1607:
Changes done in this PR:
- Refactored telemetry preferences since bugs were there in the earlier implementation.
- Updated copyright headers.
- Added telemetry tag so that users can find in settings using telemetry tag.
- Fixed some bugs around telemetry reporter.
- Telemetry disable and enable uniformed across setting changes.
- Added some unit tests

### DIFF
--- a/vscode/src/telemetry/impl/telemetryEventQueue.ts
+++ b/vscode/src/telemetry/impl/telemetryEventQueue.ts
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2024, Oracle and/or its affiliates.
+  Copyright (c) 2024-2025, Oracle and/or its affiliates.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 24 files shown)


vscode/src/telemetry/impl/telemetryEventQueue.ts
```diff
@@ -35,4 +36,26 @@ export class TelemetryEventQueue {
     this.events = [];
     return queue;
   }
+
+  public decreaseSizeOnMaxOverflow = (maxNumberOfEventsToBeKept: number) => {
```
sid-srini, line 40: The invocation of this method is passing the value of `numberOfEventsToBeDropped` as `maxNumberOfEventsToBeKept`. Although at present this value is always `capacity/2` which is same either way, it maybe better to rename either the invocation or the parameter to be consistent.

sid-srini (approved):
Thanks @Achal1607. LGTM 👍

sid-srini (approved):

PR #407, opened 2025-04-24
Updated telemetry jdk-features cache expiry time and maintenance as a

sid-srini:
- Updated telemetry jdk-features cache expiry default time to 12 hours instead of 1 hour. 
- Also, updated the jdk-features cache maintenance as a union across the source root instead of only those in the active editor.
- Added .vscode/ to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build/
 .DS_STORE
+.vscode/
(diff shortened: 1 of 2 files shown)


Achal1607 (approved):
LGTM. Thanks!

PR #408, opened 2025-04-24
Patch for the type name issue in change method parameter refactoring code action.

shivam71:
- `ChangeMethodParametersRefactoring` code action is used for changing the method signature .
- This patch fixes the issue of using fully qualified names for the parameter types irrespective of what the user has written in the code .
- The issue lead to annoying change of all the type names to fully qualified names whenever the `ChangeMethodParametersRefactoring` code action was triggered . Like `String` was converted to `java.lang.String`.
- With this patch being applied the type name would be taken from the user code as is (verbatim) .  So if the user has used simple type names they would be preserved .
- Thanks to @lahodaj for providing this patch.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -65,6 +65,7 @@
             patches/nbjavac-not-required.diff
             patches/l10n-licence.diff
             patches/dev-dependency-licenses.diff
+            patches/change-method-parameters-refactoring-qualified-names.diff
         </string>
         <filterchain>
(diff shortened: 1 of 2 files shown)


sid-srini (approved):
Thanks @shivam71 and @lahodaj. LGTM 👍

Achal1607 (approved):
LGTM. Thanks

PR #409, opened 2025-04-24
Updating telemetry consent messages

sid-srini:
1. Added a detailed data usage policy document as vscode/TELEMETRY.md
2. Updated telemetry consent settings and pop-up messages to refer to this document.


vscode/package.json
```diff
@@ -245,9 +245,10 @@
 				"jdk.telemetry.enabled": {
 					"type": "boolean",
 					"description": "%jdk.configuration.telemetry.enabled.description%",
```
Achal1607, line 247: Do we need to mention both description and markdownDescription, which is rendered where?

sid-srini: I suppose this leaves the option to the IDE to pick as per its platform limitations. In VSCode desktop env, the `markdownDescription` is picked for rendering.

sid-srini: `markdownDescription` is a VSCode specific extension as given in [their guide](https://code.visualstudio.com/docs/languages/json#_use-rich-formatting-in-hovers)

Achal1607: Thanks for the clarification

Achal1607 (approved):
LGTM. Thanks

PR #410, opened 2025-04-24
Updated messages for the minimum version of JDK to 24 for disabling nb-javac

sid-srini:
The NB25 backend includes version 24 for nb-javac. The settings for disabling nb-javac (and error messages) are updated to inform that javac from JDK24+ is needed then.


Achal1607:
LGTM. Thanks for making this change.

Achal1607 (approved):

PR #414, opened 2025-04-28
Handling class cast exception while opening pom.xml file

Achal1607:
Backport patch from netbeans which handles class cast exception while opening pom.xml in a better way.
For more info: https://github.com/apache/netbeans/pull/8460


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #416, opened 2025-04-30
Patch for changes reg. mainSource variable in Netbeans

subhash-arabhi:
Ref. [#8470](https://github.com/apache/netbeans/pull/8470)


sid-srini (changes requested):
Thanks @subhash-arabhi. Please rename the patch file as a draft and add it in sorted order in build.xml.

sid-srini (approved):
LGTM 👍

PR #418, opened 2025-04-30
added user configuration for setting up code completion commit characters

Achal1607:
- Currently `dot(".")` is provided by default as the code completion commit character alongwith `Tab` and `Enter`.
- According to VSCode guidelines and modern IDEs behaviour we should only provide `Tab` and `Enter` for accepting code completion suggestions. So, removed the default `.` character for accepting code completion.
- But, since some users might be already using `.` as their code completion character. Introduced a new configuration `jdk.java.completion.commit.chars` which accepts array of character for code completion commit.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -71,6 +71,7 @@
             patches/dev-dependency-licenses.diff
             patches/nb-telemetry.diff
             patches/change-method-parameters-refactoring-qualified-names.diff
+            patches/javavscode-375.diff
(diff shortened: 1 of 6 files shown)


sid-srini (dismissed):
Thanks @Achal1607. LGTM 👍

sid-srini (approved):
Thanks. LGTM 👍

PR #422, opened 2025-05-12
L10n 24.1.0 second set changes

naren2605:


sid-srini (approved):
Thanks @naren2605. LGTM 👍

PR #424, opened 2025-05-13
Added null pointer check for client capabilities

Achal1607:
Some of the tests on Netbeans side are facing due to null pointer exception so added an additional check


sid-srini (approved):
Thanks @Achal1607 and @shivam71. LGTM 👍

PR #425, opened 2025-05-15
Minor fix for telemetry backend and sanitisation of jdk versions

shivam71:
- Added check for https telemetry backend url and added env var to override it.

- Filtered out non alphanumeric characters from jdk versions fetched dynamically.


sid-srini (approved):
Thanks @shivam71. LGTM 👍

Issue #396, opened 2025-03-29
Formatter destroys code with syntax errors


fvclaus:
The formatter seems to have trouble handling code with syntax errors, for example:

```java
        if ("foo".contains("bar"))) {
            System.out.println("bar");
        }
```
will be formatted to 

```java
        if ("foo".contains("bar"))
            
         {
            
        }) {
            System.out.println("bar");
        }
```

This is especially annoying in larger files, where it may destroy code in the same file that is not inside your view. When you find it out, you often can't easily undo it because you might have been editing the code for a while before noticing the damage.

![Image](https://github.com/user-attachments/assets/e30d46d4-69fa-40ce-84fb-663c05c6cf5f)

anthonyvdotbe:
This is a duplicate of #195 (which itself was filed to reopen the original #65 which was closed as "non-reproducible").

PR #429, opened 2025-06-04
fix for #396 - formatting issue when there is errorneous if tree

naren2605:
as detailed in [bug#396](https://github.com/oracle/javavscode/issues/396) , [netbeans bug#8483](https://github.com/apache/netbeans/issues/8483)  and   [fix pr#8484](https://github.com/apache/netbeans/pull/8484). 

this issue happens only when user formats source file when there i an erroneous if tree. fix is provided in nb and adding patch for the same.


sid-srini (approved):
Thanks @naren2605. LGTM 👍

PR #433, opened 2025-06-17
Extension restart fix

shivam71:
**Issue**
Post activating the extension one of the two cases happen

1)`JDK Found`
2)` No JDK Found`
In case of "No JDK Found" we enter a bad state where even if the user changes the jdk path to some valid path we cannot restart unless the user reloads .

In case initial JDK found but later user changes the path to an invalid jdk home then also we enter `No JDK Found`  state.
**Fix** 
Moved the reload window pop up from jdk downloader to configuration change listener(restartExtension) triggered  whenever any configuration is changed and extension is in bad state.
This gives user a prompt to reload window so that extension can come out of the bad state.

### DIFF
--- a/vscode/src/lsp/clientPromise.ts
+++ b/vscode/src/lsp/clientPromise.ts
@@ -13,12 +13,14 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
-import { commands } from "vscode";
+import { commands,window  } from "vscode";
 import { LOGGER } from "../logger";
 import { NbProcessManager } from "./nbProcessManager";
 import { clientInit } from "./initializer";
 import { NbLanguageClient } from "./nbLanguageClient";
 import { globalState } from "../globalState";
+import { jdkDownloaderPrompt } from "../webviews/jdkDownloader/prompt";
+import { l10n } from "../localiser";
 
 export class ClientPromise {
     setClient!: [(c: NbLanguageClient) => void, (err: any) => void];
@@ -62,10 +64,15 @@ export class ClientPromise {
     public restartExtension = async (nbProcessManager: NbProcessManager | null, notifyKill: boolean) => {
         if (this.activationPending) {
             LOGGER.warn("Server activation requested repeatedly, ignoring...");
-            return;
         }
         if (!nbProcessManager) {
             LOGGER.error("Nbcode Process is null");
+            const reloadNow: string = l10n.value("jdk.downloader.message.reload");
+            const dialogBoxMessage = l10n.value("jdk.configChanged");
+            const selected = await window.showInformationMessage(dialogBoxMessage, reloadNow);
+            if (selected === reloadNow) {
+                await commands.executeCommand('workbench.action.reloadWindow');
+            }
             return;
         }
         try {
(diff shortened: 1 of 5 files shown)


vscode/src/lsp/clientPromise.ts
```diff
@@ -13,12 +13,14 @@
...
   limitations under the License.
 */
-import { commands } from "vscode";
+import { commands,window  } from "vscode";
 import { LOGGER } from "../logger";
 import { NbProcessManager } from "./nbProcessManager";
 import { clientInit } from "./initializer";
 import { NbLanguageClient } from "./nbLanguageClient";
 import { globalState } from "../globalState";
+import { jdkDownloaderPrompt } from "../webviews/jdkDownloader/prompt";
```
Achal1607: nit: remove unused import

vscode/src/lsp/clientPromise.ts
```diff
@@ -62,10 +64,15 @@ export class ClientPromise {
     public restartExtension = async (nbProcessManager: NbProcessManager | null, notifyKill: boolean) => {
         if (this.activationPending) {
             LOGGER.warn("Server activation requested repeatedly, ignoring...");
-            return;
```
Achal1607, line 65: Why this `return` statement is removed?

Achal1607 (changes requested):

vscode/src/lsp/clientPromise.ts
shivam71: 1. restartExtension gets triggered whenever the configurations are changed and in that case even if activationPending we would need to restart (as configs have changed). 
2. BTW this is linked with  issue on windows, it gets stuck here because of the return statement sometimes . So to resolve that we need to remove this return and show the reload window to the user if required.

Achal1607 (approved):
LGTM. Thank you!

sid-srini (commented):
Please rebase the branch to `main` in order to remove unrelated changes in l10n files. Thanks.

--- Update: Sorry, please ignore. I mistook the highlight due to the added `,` before the necessary changes.

shivam71:
As per the LSP protocol the client manages the [server life cycle](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lifeCycleMessages) . 

So as per the protocol the client sends a [Shutdown Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown) to the server indicating the server to prepare for shutdown . The sever must respond with errors if any that happen during the shutdown . 

On receiving the Shutdown request's response from the server the client sends an [Exit notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit) which when the server receives must kill itself i.e exit the process .

`client.stop()` ([vscode language client implementation](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L1575)) does exactly this but our netbeans lsp server doesn't implement the exit method so it keeps on running post the exit notification . Hence we have to do the mannual killing afterwards.

Implementing the `exit` method in the netbeans lsp server will help us avoid this manual killing altogether and make the code easier to understand as well as adhere with the lsp protocol spec.

PR #436, opened 2025-07-10
Fixed telemetry event issues

Achal1607:
1) Added ability to transform events while sending telemetry event.
2) Fixed possible race condition when sending JDK downloader event.


vscode/src/telemetry/utils.ts
```diff
@@ -70,3 +72,77 @@ const getUri = (pathOrUri: Uri | string): Uri => {
...
+    process.env?.LOGNAME,
+    process.env?.USER,
+    process.env?.LNAME,
+    process.env?.USERNAME,
+    process.env?.HOSTNAME,
+    process.env?.COMPUTERNAME,
+    process.env?.NAME,
+    os.userInfo().username];
+
+    return Array.from(new Set(blockedValues.filter((el): el is string => el !== undefined &&
```
sid-srini, line 92: `el` should be trimmed before checking the lengths, ignored values and adding to the Set.

vscode/src/telemetry/utils.ts
```diff
@@ -70,3 +72,77 @@ const getUri = (pathOrUri: Uri | string): Uri => {
...
+        el.length > MIN_BLOCKED_VAL_LENGTH &&
```
sid-srini:
```suggestion
        el.length >= MIN_BLOCKED_VAL_LENGTH &&
```

vscode/src/telemetry/utils.ts
```diff
@@ -70,3 +72,77 @@ const getUri = (pathOrUri: Uri | string): Uri => {
...
+    }
+    return value;
+};
+
+export const replaceAllOccurrences = (str: string, valueString: string, replaceString: string) => {
+    if(!valueString.trim().length) return str;
+    let prevString;
+    do {
+        prevString = str;
+        str = str.split(valueString).join(replaceString);
```
sid-srini: Since `string.split()` is being used, the do-while loop would not be required. That would be needed if `string.replace()` is used.

Achal1607: Sorry, my bad should have seen this.

sid-srini (approved):
Thanks @Achal1607. LGTM 👍 .
Please squash commits before merge.

Achal1607:
Squashed changes. Thanks @sid-srini

sid-srini (approved):

PR #438, opened 2025-07-15
Fixed project clean and project compile commands in the command palette

Achal1607:
The editor was not passing the necessary context to the NetBeans LS when executing the `jdk.project.run.action` command. As a result, the LS lacked the information required to determine which project or file to `compile` or `clean`, leading to an undefined error.
It worked fine with the project view because it passes the project context to the Netbeans LS.

**Fix**
This PR ensures that the appropriate context from the editor is correctly passed to the NetBeans LS, allowing it to resolve the target of the action and execute it as expected.


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #452, opened 2025-08-25
Fixed enable preview event field value for Gradle and Standalone projects

Achal1607:
Standalone issue:
There was a race condition, the Netbeans LS requested for run config options from the client but was not waiting for the response. It sent the response before the future completes, due to which it always returned `false`.

Gradle issue fixed by @sid-srini :
1. For gradle projects, the CompilerOptionsQuery provides the required compiler arguments for a Source file/folder, not the project root. Thus, fixed the public LspServerTelemetryManager.isPreviewEnabled() methods to query the source roots of the current gradle project and sub-projects in order to check for the existence of the enable-preview compiler argument.
2. Fixed LspServerTelemetryManger.getProjectType() to return "gradle" only when the project name contains "gradle". For other kinds of netbeans projects, not maven/gradle, the returned value is null. This avoids wrong metrics collected for unsupported project types, such as ant, openjdk, make etc.


sid-srini (approved):
Thanks a lot @Achal1607 for the fixxes, and, applying the gradle fix for other projects like maven too. LGTM 👍

PR #453, opened 2025-08-25
Constructor init of java.lsp.server JavaPlatformProvider override

sid-srini:
1. Brings NetBeans PR apache/netbeans#8745 as a patch

2. Allowed `LspJavaPlatformProviderOverride` to initialize the override based on the value set for the option "netbeans.lsp.java.platform.override" passed as a JVM system property.

3. Utilised platform override system property in Java VSCode extension
    1. From the Java VSCode extension, passed the value of "`jdk.project.jdkhome`" configuration to the NBLS via the JVM system property flag "`netbeans.lsp.java.platform.override`".
    2. This additionally allows the `LspJavaPlatformProviderOverride` to initialize the `JavaPlatform` override during construction, avoiding any caching of the default platform prior to the configuration being read via LSP.

Thanks to @lahodaj for the idea of this fix.

### DIFF
--- a/vscode/src/configurations/handlers.ts
+++ b/vscode/src/configurations/handlers.ts
@@ -55,6 +55,10 @@ export const jdkHomeValueHandler = (): string | null => {
         null;
 }
 
+export const projectJdkHomeValueHandler = (): string | null => {
+    return getConfigurationValue(configKeys.projectJdkHome, null);
+}
+
 export const projectSearchRootsValueHandler = (): string => {
     let projectSearchRoots: string = '';
     const isProjectFolderSearchLimited: boolean = !getConfigurationValue(configKeys.disableProjSearchLimit, false);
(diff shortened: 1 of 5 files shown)


Achal1607 (approved):
LGTM. Thanks

vscode/src/configurations/handlers.ts
```diff
@@ -55,6 +55,11 @@ export const jdkHomeValueHandler = (): string | null => {
         null;
 }
 
+export const projectJdkHomeValueHandler = (): string | null => {
+    return getConfigurationValue(configKeys.projectJdkHome) ||
```
Achal1607: nit: you can directly pass as second argument `null` to set the default value, no need to use the `||` operator.

sid-srini: Thanks. I've fixed the code accordingly.

Achal1607 (dismissed):
LGTM. Thanks for fixing this

Achal1607 (approved):

lahodaj (approved):
Looks OK to me.

PR #466, opened 2025-09-18
Fixing outline view of compact source files

shivam71:
- Compact source files in java don't have top level class.
- Methods and fields are allowed as top level elements .
- Due to the erroneous computation of the end position for the hidden final class created to scaffold the compact source file methods and fields the outline view was not showing the document symbols for these files.
- This patch does a simple fix to ensure that the LSP backend sends ranges where end position for a document symbol if not found is taken same as the start position .
- Final fix could perhaps correct the end position computation directly .

### DIFF
--- a/patches/8856-draft.diff
+++ b/patches/8856-draft.diff
@@ -1,0 +1,15 @@
+diff --git a/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementOpen.java b/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementOpen.java
+index f561cb51c8..81000c3ed9 100644
+--- a/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementOpen.java
++++ b/java/java.sourceui/src/org/netbeans/api/java/source/ui/ElementOpen.java
+@@ -573,6 +573,10 @@ public final class ElementOpen {
+         int[] span = null;
+         switch(forTree.getKind()) {
+             case CLASS:
++                if ((int) target[1] >= 0 && (int) target[2] == -1) {
++                // Compact Source file (JEP 512)  issue implicit class end position not found in code 
++                    target[2] = (int) info.getTrees().getSourcePositions().getEndPosition(cu, cu);
++                }
+             case INTERFACE:
+             case ENUM:
+             case ANNOTATION_TYPE:
--- a/build.xml
+++ b/build.xml
@@ -57,6 +57,7 @@
             patches/8745-draft.diff
             patches/8770.diff
             patches/8827.diff
+            patches/8856-draft.diff
             patches/disable-error-notification.diff
             patches/mvn-sh.diff
             patches/project-marker-jdk.diff


patches/8856-draft.diff
sid-srini, line 1: It seems that this class already does perform such a check in the public method `getLocation`: lines 327-334. 

It utilizes the approach that if the end is not found (-1), set it to the end of the source file. Similarly, if the start is not found (-1), set it to the start of the source file. 

There may be a reason why the absent values are not corrected at the source itself, and done in specific downstream usages.

So, perhaps the `getLocation` method should be used if not used, or, the logic used in `getLocation` should be copied in the places where `getOpenInfo` and `getFutureOpenInfo` are being used directly. For example, in its usages in `LspElementUtils`, the method `fillInPositions` uses the same logic to update the absent positions.

sid-srini (commented):
Thanks @shivam71 for investigating this issue. Some suggestions on the temp fix are below.

patches/8856-draft.diff
shivam71: Thanks @sid-srini ! 
I spend a bit more time understanding how the flow goes .  

- For a java element present in the document its source may be part of the same file or the element may be coming from something on the class path 
- In case the element is not in the current file (the current compilation unit) then we utilise the `getLocation` method (uses the class path)
- Further `fillInTreePositions` method is a downstream method used by the different `getOpenInfo` methods that are used by `getLocation` 
- As far as the checks performed by the `getLocation` and `fillInPositions` methods they fall back on the expanded range computed when the selected range is not valid .
Like in the following snippet part of the `getLocation` method
```java                
int start = (int) openInfo[3];   
if (start < 0) {
    start = Math.max(0, (int) openInfo[1]); // falling back here 
}
int end = (int) openInfo[4];
if (end < 0) {
    end = Math.max(0, (int) openInfo[2]); // falling back here 
 }
```

- In the `fillInTreePositions` method we need to correct the expanded range ( first pair at index 1 and 2) (openInfo comes from  `fillInTreePositions` eventually )
- From the compilation unit tree we can find the length of the source file and can utilise that instead of making the end equal to the start position as in `cu.getSourceFile().getCharContent(true).length()` (when the getSourceFile doesn't return null)
So as far as I understand the initial pair of the expanded range need to be corrected in `fillInTreePositions` method  itself while the upstream methods can correct the remaining exact selection range by falling back on these ones . I am not very sure of this so please correct wherever necessary .

sid-srini: Thanks. So then only the first fix i.e. of `target[2]`, and potentially of `target[1]`, if `< -1` needs to be done in `fillInTreePositions`? The downstream methods then can continue to handle the element positions in the way they want to.
Also, in most cases, the source info would already be pre-populated in `target[1]` and `target[2]`. Is there a need to check the logic of this further upstream?

shivam71: Yes, once we fix the source info then upstream checks may not be required .

sid-srini (dismissed):
Thanks a lot @shivam71 for further refining the fix. LGTM 👍

sid-srini:
> Netbeans PR [#8856](https://github.com/apache/netbeans/pull/8856)

nit: would you please rename the patch file to `8856-draft.diff` so that we may track it better when integrating a future NB milestone. Thanks

sid-srini (approved):

PR #467, opened 2025-09-19
Fix parameter field buttons in Change Method Parameters UI in Language Server

Achal1607:
Back-ported patch from NB-28. 
Check for more details: https://github.com/apache/netbeans/pull/8829


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #487, opened 2025-10-14
Backporting Support running classes from specific Gradle source sets in Gradle actions

Achal1607:
More detailed info, please refer: https://github.com/apache/netbeans/pull/8897


sid-srini (dismissed):

PR #494, opened 2025-10-28
Refactored file URI usages to use a common utility method

Achal1607:
Refactored file URI usages to use a common utility method for consistent handling and to avoid ambiguity between `Uri.file` and `Uri.parse`.


sid-srini (approved):

PR #496, opened 2025-10-28
Updated LSP new from template command input validation rules

shivam71:
1. Allowed 'module-info' and 'package-info' as special cases.
2. Allowed only simple names which are not keywords hence not allowing names which have dots.


sid-srini (approved):
Thanks @shivam71. LGTM 👍

PR #498, opened 2025-10-29
Fixed URI issue while reading project context mapping for notebooks

Achal1607:
Fixes an issue where project context lookup failed because the URI path was used instead of the OS specific absolute path stored in the map.
The fix ensures the URI is first converted to an absolute path before lookup, allowing correct project context resolution across platforms.


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #501, opened 2025-10-31
Handle user dismissal of input box in notebook instead of closing input stream

Achal1607:
JShell automatically closes the input stream when `-1` is returned via `CustomInputStream`, with no mechanism to reopen or reset it.
This change replaces `-1` returns with `EOFException` to safely signal user dismissal while keeping the stream open.


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #504, opened 2025-11-04
remove l10n lib dependency

naren2605:
remove l10n lib dependency and use format method in  code base from l10n library repo

### DIFF
--- a/vscode/src/localiser.ts
+++ b/vscode/src/localiser.ts
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2023-2024, Oracle and/or its affiliates.
+  Copyright (c) 2023-2025, Oracle and/or its affiliates.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


vscode/src/localiser.ts
```diff
@@ -34,21 +32,21 @@ export interface l10n {
...
       const isPresentInBundle = valueFromBundle !== key;
       return isPresentInBundle ? valueFromBundle : this.defaultTranslation(key, placeholderMap);
     }
+    defaultTranslation(key: string, placeholderMap:Record<string, any>):string{
+          let value = this.defaultl10nContent[key];
+          return value?this.format(value,placeholderMap):key;
+    }
     nbLocaleCode(){
       const vscodeLanguage = vscode.env.language;
       if (!vscodeLanguage) return DEFAULT_LANGAUGE;
```
Achal1607: nit: typo `DEFAULT_LANGAUGE` -> `DEFAULT_LANGUAGE`
This line is not changed in this PR but typo maybe from earlier changes.

vscode/src/localiser.ts
```diff
@@ -34,21 +32,21 @@ export interface l10n {
...
-  private defaultTranslation: TranslatorFn;
-
+    private defaultl10nContent:any;
     constructor(extensionId: string, defaultBundlePath: string) {
       let defaultBundleAbsoluteFsPath = FileUtils.toUri(`${vscode.extensions.getExtension(extensionId)?.extensionPath}/${defaultBundlePath}`).fsPath;
-      l10nLib.config({
-        fsPath: defaultBundleAbsoluteFsPath
-      });
-      this.defaultTranslation = l10nLib.t;
+      this.defaultl10nContent = JSON.parse(fs.readFileSync(defaultBundleAbsoluteFsPath,'utf-8'));
```
Achal1607: Also, should we have a `try-catch` incase the file is not found or there is some issue in the JSON parsing, or that case is never possible?

naren2605: we can just add log line with error and we stil have to throw the exception, as this file is mandatory while initializing

Achal1607: Yes, maybe we can do that then so that at least from the logs we can get error info and a generic exception can be re-thrown.

vscode/src/localiser.ts
```diff
@@ -16,49 +16,104 @@
...
-      let defaultBundleAbsoluteFsPath = FileUtils.toUri(`${vscode.extensions.getExtension(extensionId)?.extensionPath}/${defaultBundlePath}`).fsPath;
-      l10nLib.config({
-        fsPath: defaultBundleAbsoluteFsPath
-      });
-      this.defaultTranslation = l10nLib.t;
+  private defaultl10nMap: any;
+  constructor(extensionId: string, defaultBundlePath: string) {
+    let extnPath = vscode.extensions.getExtension(extensionId)?.extensionPath;
+    let defaultBundleAbsoluteFsPath = FileUtils.toUri(`${extnPath}/${defaultBundlePath}`).fsPath;
+    let logAndThrowError = (message: string, err: unknown) => {
```
Achal1607: Can we move this to utils so that it can be reused in future as well?

vscode/src/localiser.ts
```diff
@@ -16,49 +16,104 @@
...
 import { extConstants } from './constants';
 import { FileUtils } from './utils';
+import { LOGGER } from './logger';
 
-const DEFAULT_LANGAUGE = "en";
-const DEFAULT_BUNDLE_FILE = `l10n/bundle.l10n.${DEFAULT_LANGAUGE}.json`;
-
-type TranslatorFn = typeof vscode.l10n.t
+const DEFAULT_LANGUAGE = "en";
+const DEFAULT_BUNDLE_FILE = `l10n/bundle.l10n.${DEFAULT_LANGUAGE}.json`;
```
Achal1607: Nit: We were already using the format `l10n/bundle.l10n.${DEFAULT_LANGUAGE}.json`, but we can switch to `path.join` if you think it makes things cleaner.

vscode/src/test/unit/mocks/vscode/mockVscode.ts
Achal1607: nit: We can also update the copyright from `2023-2024` to `2023-2025`

vscode/src/test/unit/mocks/init.ts
Achal1607, line 1: nit: Same here, we can update the copyright

vscode/src/localiser.ts
Achal1607, line 1: nit: Same here update the copyright

Achal1607 (changes requested):

vscode/src/localiser.ts
```diff
@@ -16,49 +16,97 @@
...
 'use strict';
 
-import * as l10nLib from '@vscode/l10n'
-
 import * as vscode from 'vscode';
+import * as fs from 'fs';
 import { extConstants } from './constants';
 import { FileUtils } from './utils';
+import { LOGGER } from './logger';
+import path = require('path');
```
Achal1607: nit: `import path = require('path');` can be written as `import * as path from 'path';` for better consistency.

Achal1607 (dismissed):
nit: Please format `localisation/extension.test.ts` file if not already done.
Also, please squash the commits at the end.

PR #508, opened 2025-11-06
Set user.dir for JShell remote VM to notebook's parent directory

Achal1607:
This PR configures the JShell remote VM's `user.dir` system property to point to the notebook file's parent directory, ensuring that relative file paths in notebook code cells are resolved relative to the notebook's location rather than the Netbean's user directory.

### DIFF
--- a/nbcode/notebooks/src/org/netbeans/modules/nbcode/java/notebook/NotebookSessionManager.java
+++ b/nbcode/notebooks/src/org/netbeans/modules/nbcode/java/notebook/NotebookSessionManager.java
@@ -18,6 +18,8 @@ package org.netbeans.modules.nbcode.java.notebook;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import java.net.URI;
+import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


nbcode/notebooks/src/org/netbeans/modules/nbcode/java/notebook/NotebookSessionManager.java
```diff
@@ -99,11 +101,20 @@ private JShell jshellBuildWithProject(Project prj, JshellStreamsHandler streamsH
...
                     .remoteVMOptions(remoteOptions.toArray(new String[0]));
+        } else if (!remoteOptions.isEmpty()) {
+            builder.remoteVMOptions(remoteOptions.toArray(new String[0]));
         }
 
         return builder.build();
     }
 
+    private void setSystemPropertiesForRemoteVm(List<String> remoteOptions, String notebookUri) {
+        String parentPath = Path.of(URI.create(notebookUri)).getParent().toString();
```
sid-srini: Perhaps some try-catch and error handling should be added here. Further, the parent Path should be checked to see if it is a directory or not and only set the property if valid.

sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #509, opened 2025-11-07
Updating file name input validation rule to give warning for dollar characters

shivam71:
Dollar characters are allowed for java objects but are not recommended for user defined objects this adds a check for the same .

### DIFF
--- a/patches/7893-draft.diff
+++ b/patches/7893-draft.diff
@@ -1,5 +1,5 @@
 diff --git a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
-index 891bfa328d..91172fda4b 100644
+index 891bfa328d..e7359322b6 100644
 --- a/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
 +++ b/java/java.lsp.server/src/org/netbeans/modules/java/lsp/server/protocol/LspTemplateUI.java
 @@ -27,6 +27,7 @@ import java.net.MalformedURLException;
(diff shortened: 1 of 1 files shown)


patches/7893-draft.diff
```diff
@@ -119,6 +124,18 @@ index 891bfa328d..91172fda4b 100644
 +                    {
 +                        client.showMessage(new MessageParams(MessageType.Error,Bundle.ERR_InvalidObjectName(name)));
 +                        return client.showInputBox(new ShowInputBoxParams(Bundle.CTL_TemplateUI_SelectName(), desc.getProposedName())).thenCompose(this);
++                    } else if (name.contains("$")) {
++                        final MessageActionItem continue_action = new MessageActionItem("Continue");
```
Achal1607: Please add it in the bundle so that it can be localized.

patches/7893-draft.diff
```diff
@@ -119,6 +124,18 @@ index 891bfa328d..91172fda4b 100644
...
++                        final MessageActionItem try_again = new MessageActionItem("Try Again");
```
Achal1607: Please add it in the bundle so that it can be localized.

patches/7893-draft.diff
```diff
@@ -119,6 +124,18 @@ index 891bfa328d..91172fda4b 100644
...
++                        ShowMessageRequestParams smrp = new ShowMessageRequestParams(Arrays.asList(continue_action, try_again));
++                        smrp.setMessage("Object name contains $, not recommended for user defined objects , Do you want to continue with the current name or try again ?");
```
Achal1607: Please add it in the bundle so that it can be localized.

Achal1607 (changes requested):

patches/7893-draft.diff
sid-srini: nit: Try to use the camel-case codestyle for naming variables etc.

patches/7893-draft.diff
```diff
@@ -27,6 +32,10 @@ index 891bfa328d..91172fda4b 100644
      "CTL_TemplateUI_SelectPackageName=Package name of your project?",
      "CTL_TemplateUI_SelectPackageNameSuggestion=org.yourcompany.yourproject",
      "CTL_TemplateUI_SelectName=Name of the object?",
++    "PROMPT_AskObjectNameWarningConfirmation=Object name contains $. This not recommended for user defined objects. "
```
Achal1607: `This not recommended` should be replaced by `This is not recommended` and space before `?` should be removed.
Also, maybe `$` can be replaced with `'$'`, if it makes sense.

Achal1607 (changes requested):

patches/7893-draft.diff
shivam71: Thanks ! done

PR #515, opened 2025-11-12
Introduce open extension user directory command and improve cache deletion error prompt

shivam71:
* Introduce open extension user directory command.
   *  Opens the language server user directory for the current workspace
* Improve delete cache error to prompt for manual deletion.
   * Added button to open the language server user directory


vscode/src/commands/cache.ts
```diff
@@ -13,23 +13,25 @@
...
 import { ICommand } from "./types";
 import { l10n } from "../localiser";
 import * as fs from 'fs';
 import * as path from 'path';
 import { globalState } from "../globalState";
 
-const deleteCache = async () => {
-    // TODO: Change workspace path to userdir path
+const getLanguageServerUserDir = () => {
     const storagePath = globalState.getExtensionContextInfo().getWorkspaceStorage()?.fsPath;
```
Achal1607: Storage path depends on the `user_dir` configuration so it be should be, which is also mentioned in the `TODO` comment above:
```
  const userdirScope = process.env['nbcode_userdir'] || getConfigurationValue(configKeys.userdir, "local");
    const workspaceStoragePath = extensionContextInfo.getWorkspaceStorage()?.fsPath;
    const userdirParentDir = userdirScope === "local" && workspaceStoragePath
        ? workspaceStoragePath
        : extensionContextInfo.getGlobalStorage().fsPath;
```

vscode/src/commands/cache.ts
```diff
@@ -44,7 +46,14 @@ const deleteCache = async () => {
                 await fs.promises.rm(userDir, { recursive: true });
                 await window.showInformationMessage(l10n.value("jdk.extension.message.cacheDeleted"), reloadWindowActionLabel);
             } catch (err) {
-                await window.showErrorMessage(l10n.value("jdk.extension.error_msg.cacheDeletionError"), reloadWindowActionLabel);
+                console.log(err);
```
Achal1607: nit: Please use `LOGGER` instead of logging into `console`

vscode/src/commands/cache.ts
```diff
@@ -44,7 +46,14 @@ const deleteCache = async () => {
...
+                const openLSUserDirLabel = l10n.value("jdk.extension.cache.label.openLSUserDir");
+                const selectedAction = await window.showErrorMessage(l10n.value("jdk.extension.error_msg.cacheDeletionError", { userDir }), openLSUserDirLabel);
+                if (selectedAction === openLSUserDirLabel) {
+                    const opened = await openLanguageServerUserDir();
+                    if (opened)
+                        await window.showErrorMessage(l10n.value("jdk.extension.cache.message.reloadWindow.afterUserDirDeletion"), reloadWindowActionLabel);
```
Achal1607: here `reloadWindowActionLabel` doesn't make any difference, like even if the user doesn't ask for reload it would reload anyhow?

vscode/src/commands/cache.ts
```diff
@@ -53,8 +62,22 @@ const deleteCache = async () => {
         window.showErrorMessage(l10n.value("jdk.extension.cache.message.noUserDir"));
     }
 }
+const openLanguageServerUserDir = async () => {
+    const userDir = getLanguageServerUserDir();
+    if (userDir && fs.existsSync(userDir)) {
+       return env.openExternal(Uri.file(userDir));
```
Achal1607: Instead of `Uri.file` please use `FileUtils.toUri` method.

vscode/src/commands/cache.ts
Achal1607: nit: Please update the copyright from `2023-2024` to `2023-2025`

vscode/src/commands/cache.ts
shivam71: Hmm, You suggest we remove it ?

Achal1607: @sid-srini please suggest the intended behaviour here

vscode/src/test/integration/suite/general/extension.test.ts
```diff
@@ -306,5 +306,10 @@ suite('Extension Test Suite', function () {
     const formattedContents = editor.document.getText().trim();
     assert.ok(formattedContents == FORMATTED_POM_XML.trim(), "pom.xml is not formatted");
   });
-
+  // Check if open  language server  user directory command is excuted succesfully
```
Achal1607: we can add unit tests for testing of the business logic instead of integration tests, as this test doesn't help much in testing business logic in this case.

Achal1607 (changes requested):

PR #524, opened 2025-11-25
Update notebook samples

shivam71:
* Made current sample use JDK 25 IO APIs
* Moved previous sample notebook to `LearningSample_Java17+.ijnb`


sid-srini (approved):
Thanks @shivam71. LGTM 👍

PR #541, opened 2026-01-12
Document service refactoring  changes

shivam71:


patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
++                            Class<?> paramClass = p.getClass();
++                            Method getTextDoc = paramClass.getMethod("getTextDocument");
++                            TextDocumentIdentifier textDocId = (TextDocumentIdentifier)getTextDoc.invoke(p);
++                            String uri = textDocId.getUri();
++                            if(uri.startsWith(NOTEBOOK_TEXT_DOC_URI_IDENTIFIER)){
++                                String newMethod = rq.getMethod().replaceFirst("textDocument", "notebookDocument");
++                                rq.setMethod(newMethod);
++                            }
++                        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException | NullPointerException ex) {
++
```
Achal1607, line 724: What would be the behaviour of the request if a notebook sends a request for let's day formatter but if that method is not implemented by the notebook then would it come in the catch block?
If yes, then can we add at least a `FINE` level logger so that exceptions are not swallowed.

patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
  import org.netbeans.modules.java.lsp.server.progress.OperationContext;
  import org.netbeans.modules.parsing.spi.indexing.Context;
  import org.netbeans.modules.parsing.spi.indexing.CustomIndexer;
-@@ -1383,6 +1386,16 @@ public final class Server {
+@@ -253,7 +257,7 @@ public final class Server {
+         private NbCodeLanguageClient client;
+         private OperationContext initialContext;
+         private List<Object> additionalServices = new ArrayList<>();
+-
++        private static final String NOTEBOOK_TEXT_DOC_URI_IDENTIFIER = "vscode-notebook-cell";// NOI18N
```
Achal1607: If you think `vscode-notebook-cell` is a property of NotebookDocumentService, then you can create a static member in the `NotebookDocumentService` and then reference it everywhere for identifier.

patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
++                        try {
++                            Class<?> paramClass = p.getClass();
++                            Method getTextDoc = paramClass.getMethod("getTextDocument");
++                            TextDocumentIdentifier textDocId = (TextDocumentIdentifier)getTextDoc.invoke(p);
++                            String uri = textDocId.getUri();
++                            if(uri.startsWith(NOTEBOOK_TEXT_DOC_URI_IDENTIFIER)){
++                                String newMethod = rq.getMethod().replaceFirst("textDocument", "notebookDocument");
++                                rq.setMethod(newMethod);
++                            }
++                        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException | NullPointerException ex) {
```
Achal1607, line 723: What would be the behaviour of the request if a notebook sends a request for let's day formatter but if that method is not implemented by the notebook then would it come in the catch block?
Also, can we add at least a `FINE` level logger in the catch block so that exceptions are not swallowed.

patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
+             this.sessionLookup = sessionLookup;
+         }
+@@ -311,6 +315,21 @@ public final class Server {
+                         if (p instanceof WorkDoneProgressParams) {
+                             ctx.setProgressToken(((WorkDoneProgressParams)p).getWorkDoneToken());
+                         }
++                        // LSP Client sends requests for notebook cells with textDocument prefix
++                        // Our notebookDocumentService implementation requires different implementation for these requests
++                        try {
++                            Class<?> paramClass = p.getClass();
```
Achal1607, line 715: Is there any alternate approach possible here, apart from using reflection?

patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
++                        }        
++                        
+                     } else if (msg instanceof NotificationMessage) {
+                         NotificationMessage not = (NotificationMessage)msg;
+                         Object p = not.getParams();
+@@ -412,6 +431,7 @@ public final class Server {
+         private static final Logger LOG = Logger.getLogger(LanguageServerImpl.class.getName());
+         private NbCodeClientWrapper client;
+         private final TextDocumentServiceImpl textDocumentService = new TextDocumentServiceImpl(this);
++        private final NotebookDocumentServiceHandler notebookDocService = Lookup.getDefault().lookup(NotebookDocumentServiceHandler.class);
```
Achal1607, line 734: Perhaps `Lookup.getDefault().lookup(NotebookDocumentServiceHandler.class)` can be moved into the `getNotebookDocumentService()` method so that the lookup is performed only when needed. Additionally, if `NotebookDocumentServiceHandler` is not yet loaded during `Server` class initialization, the current approach could lead to issues.

patches/java-notebooks.diff
```diff
@@ -649,79 +611,149 @@
...
++            return notebookDocService;
++        }
+         @Override
+         public WorkspaceService getWorkspaceService() {
+             return workspaceService;
+@@ -1154,6 +1177,7 @@ public final class Server {
+             sessionServices.add(inputService.getRegistry());
+             sessionServices.add(workspaceService.getWorkspace());
+             ((LanguageClientAware) getTextDocumentService()).connect(client);
++            ((LanguageClientAware) getNotebookDocumentService()).connect(client);
```
Achal1607: A `null` check for `getNotebookDocumentService()` would be good, since we don't contribute our `nbcode` to Netbeans, but if we contribute the LSP code to Netbeans, it will not find any `NotebookDocumentServiceHandler` and fail with a NPE before starting the server itself.

patches/java-notebooks.diff
```diff
@@ -765,441 +807,368 @@ index 747d151600..bbf1f263d1 100644
...
+-
+     @ServiceProvider(service=IndexingAware.class, position=0)
+     public static final class RefreshDocument implements IndexingAware {
+ 
+@@ -601,34 +598,71 @@ public class TextDocumentServiceImpl implements TextDocumentService, LanguageCli
+     }
+ 
+     public void init(ClientCapabilities clientCapabilities, ServerCapabilities severCapabilities) {
++        setServerSemanticTokenProvider(clientCapabilities,severCapabilities); 
++        setServerNotebookCaps(severCapabilities);
```
Achal1607: Since there is now proper isolation between the `TextDocumentService` and the `NotebookDocumentService`, would it be possible to move `setServerNotebookCaps` into the `NotebookDocumentService` as well?
I’m not entirely sure whether this is feasible, it’s just a thought, but if it is, it would be good to move this responsibility out as well for better separation of concerns.

patches/java-notebooks.diff
```diff
@@ -765,441 +807,368 @@ index 747d151600..bbf1f263d1 100644
...
++import org.netbeans.modules.java.lsp.server.notebook.NotebookDocumentServiceHandler;
+ import org.netbeans.modules.java.lsp.server.refactoring.ChangeMethodParameterUI;
+ import org.netbeans.modules.java.lsp.server.refactoring.MoveElementUI;
+ import org.netbeans.modules.java.lsp.server.refactoring.ParameterUI;
+@@ -391,7 +395,24 @@ public class ServerTest extends NbTestCase {
              return CompletableFuture.completedFuture(null);
+         }
+     }
+-
++    public void testDelegationForNotebookCells() throws Exception {
```
Achal1607, line 1158: Can we add one more test, for a method which is not implemented, so that we know that the exception is thrown and caught in the catch block?

nbcode/notebooks/src/org/netbeans/modules/nbcode/java/notebook/NotebookDocumentServiceHandlerImpl.java
Achal1607: Also, add copyright `2025-2026`

patches/java-notebooks.diff
shivam71: Yes if the method is not implemented we get a JSON-RPC error of UnsupportedRequestMethod exception . 
In this try catch however that cannot be caught , here the reflection based errors are being caught . Hence that exception is not being swallowed here .

patches/java-notebooks.diff
shivam71: Sure !

patches/java-notebooks.diff
shivam71: Core idea was to make it generic so in future we can just add new methods in NotebookDocumentServiceHandler
without making any changes in this part of the code . For now we could have just done this delegation for completion request in particular and not do any reflection .

shivam71: Checked for using instanceOf approach . Super type `TextDocumentPosition` is applicable to handful of the methods . There are many methods which do not have any common super type but they have document uri being send .  For example 

1. `CodeActionParams`
2. `CodeLensParams`
3. `DocumentFormattingParams`
4. `DocumentSymbolParams`
5. `FoldingRangeRequestParams`
6. `InLayHintsParams`
and some more .
Given this it seems to may that doing `instanceOf` for each is not suitable . 
I suggest we go with reflection for now and later we can perhaps change to method handles if required .

PR #549, opened 2026-02-04
Fixed unit tests when run on windows

sid-srini:


vscode/src/test/unit/general/localiser.unit.test.ts
```diff
@@ -18,19 +18,21 @@ import { describe, it, beforeEach, before, after } from "mocha";
...
 
 describe('localiser tests', () => {
+    const slash: string = os.type() === 'Windows_NT' ? '\\' : '/';
     let loggerLogStub: sinon.SinonStub;
     let mockedExtns: typeof vscode.extensions;
     let extMock: Extension<any>;
     let mockedEnv: typeof vscode.env;
     let mockedL10n: typeof vscode.l10n;
-    let currentDir = __dirname.replace("/out/", "/src/");
+    let currentDir = __dirname.replace(slash + "out" + slash, slash + "src" + slash);
```
Achal1607: is there any method in `os` or `path` package that we can use, so that it is easier for maintainence?

sid-srini: Thanks for the pointer. I have updated it to now use `path.sep` instead.

Issue #64, opened 2023-11-04
Add support for inline breakpoints for lambda


gayanper:
When debugging a line with lambda, in vscode terms you can use inline breakpoints just to suspend at a specific lambda. Today this extension only supports line breakpoints and inline breakpoints added in vscode ui side are ignored and treated as line breakpoints.

gayanper:
![](https://user-images.githubusercontent.com/14052197/190965636-d2794c99-115e-4d72-8269-b7f6e9d38efe.png)

gayanper:
@pron As you requested in #31

gayanper:
@Achal1607 any update from the NB side for this ?

gayanper:
Any update on this issue ?

PR #571, opened 2026-03-18
Breakpoints for lambdas.

lahodaj:
This is a draft/prototype incorporating https://github.com/apache/netbeans/pull/9285 into the extension, for testing purposes. It adds an ability to put breakpoints in individual lambdas. Internally, the lambdas are identified by the sequence number on their (starting) line. Please look at the NB PR for some details.

In VS Code, it may look like:
vscode-lambda-breakpoints-editor-1
vscode-lambda-breakpoints-editor-2

Please see also https://github.com/oracle/javavscode/issues/64.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -74,6 +74,7 @@
             patches/updated-show-input-params.diff
             patches/java-notebooks.diff
             patches/formatOptions.diff
+            patches/9285.diff
         </string>
         <filterchain>
             <tokenfilter delimoutput=" ">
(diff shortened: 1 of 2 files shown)


sid-srini (commented):
Thanks a lot @lahodaj. This is a much needed feature. 

One minor nit: please rename the patch file to something like 9285-draft.diff so that we may easily track the upstream PR. Thanks.

Achal1607 (commented):
Thanks @lahodaj,  I’ve done an overview review, and it looks good to me as an initial patch. This is an important feature.

sid-srini (approved):
Thanks a lot @lahodaj. LGTM 👍

PR #585, opened 2026-04-21
Fixed ordering of commands in the explorer window

### LINKED ISSUE #564: Annoying directory context menu options
When the Java extension is installed / enabled and I right-click a directory in the explorer, I see "Create New Notebook…" and "New File from Template…" as the first two options, before "New File…" and "New Folder…" which are usually the first two. This is very annoying as I have developed muscle memory for creating new directories by right clicking, moving the mouse to the second entry and clicking it.

Could you please have these items further down or allow me to disable them through configuration?

Achal1607:
We had earlier set our commands to start from position `3`, but it seems that VS Code’s native commands use the ordering `4` and `6`. So, for our commands to come after those, we need to start our ordering from `7`. This has now been fixed.

Updated Explorer window context:
Screenshot 2026-04-21 at 3 06 45 PM (2)

closes #564

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -763,12 +763,12 @@
 				{
 					"command": "jdk.workspace.new",
 					"when": "nbJdkReady && explorerResourceIsFolder",
-					"group": "navigation@3"
+					"group": "navigation@7"
 				},
 				{
 					"command": "jdk.notebook.new",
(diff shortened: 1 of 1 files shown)


sid-srini (approved):
Thanks @Achal1607. LGTM 👍

PR #586, opened 2026-04-23
Relaxed version check on the notebook format

Achal1607:
Currently, there is a version validation check on the notebook version: it must be greater than or equal to 4.5, and the major version must be 4.

This requirement is now being relaxed to allow versions greater than or equal to 4.5 with a major version above 4 as well. This change ensures that users with newer notebook versions can still use our extension, as long as their notebooks adhere to the 4.5 schema and do not introduce any breaking changes.

### DIFF
--- a/vscode/src/notebooks/nbformat.v4.d7.schema.json
+++ b/vscode/src/notebooks/nbformat.v4.d7.schema.json
@@ -94,15 +94,14 @@
       }
     },
     "nbformat_minor": {
-      "description": "Notebook format (minor number). Incremented for backward compatible changes to the notebook format.",
+      "description": "Notebook format (minor number). Incremented for backward compatible changes to the notebook format. Note: the minimum version constraint has been relaxed from the original spec (5 → 0) to allow validation of newer major versions beyond those explicitly defined.",
       "type": "integer",
-      "minimum": 5
+      "minimum": 0
     },
     "nbformat": {
       "description": "Notebook format (major number). Incremented between backwards incompatible changes to the notebook format.",
       "type": "integer",
-      "minimum": 4,
-      "maximum": 4
+      "minimum": 4
     },
     "cells": {
       "description": "Array of cells of the current notebook.",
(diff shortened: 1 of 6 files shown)


vscode/src/notebooks/notebook.ts
```diff
@@ -91,6 +91,18 @@ export class Notebook {
             LOGGER.error(`Notebook JSON validation failed:\n${errors}`);
             throw new Error(l10n.value("jdk.notebook.validation.failed.error_msg"));
         }
+
+        if (!Notebook.isSupportedNotebookVersion(notebook)) {
+            LOGGER.error(`Notebook version validation failed: unsupported notebook version ${notebook.nbformat}.${notebook.nbformat_minor}.`);
+            throw new Error(l10n.value("jdk.notebook.validation.failed.error_msg"));
```
sid-srini: nit: It maybe useful to log or/and display in the error message which is the minimum supported version i.e. 
```js
`${NotebookVersionInfo.NBFORMAT}.${NotebookVersionInfo.NBFORMAT_MINOR}`
```

vscode/src/notebooks/nbformat.v4.d7.schema.json
```diff
@@ -96,13 +96,12 @@
     "nbformat_minor": {
       "description": "Notebook format (minor number). Incremented for backward compatible changes to the notebook format.",
```
sid-srini: nit: It maybe useful to add a suffixed note in the description that the minimum version has been modified from the spec: 5 &rarr; 0 due to relaxation to support higher major versions.

Perhaps in the future, we could maintain the schema spec for each major version that is supported and conditionally validate against it.

sid-srini (approved):
Thanks @Achal1607. LGTM 👍 

Minor comments below are mostly for future reference. Thanks.

vscode/src/notebooks/notebook.ts
Achal1607: Thanks, added. Please review once if the message looks fine

vscode/src/notebooks/nbformat.v4.d7.schema.json
Achal1607: Done. Thanks

sid-srini (approved):

PR #587, opened 2026-04-24
Updated downloader constant for openjdk base url from 26 to 26.0.1

sid-srini:


Achal1607 (approved):
LGTM. Thanks for the update @sid-srini

PR #590, opened 2026-04-24
Updated sbom generation tools patch versions

sid-srini:


Achal1607 (approved):
LGTM. Thanks @sid-srini

Issue #589, opened 2026-04-24
javac-related AssertionError with JDK 26


anthonyvdotbe:
extension version: 25.1.0

extension configuration:

```plaintext
"jdk.advanced.disable.nbjavac": true,
"jdk.jdkhome": "C:\\dev\\jdk-26",
```

stack trace:

```plaintext
SEVERE [org.openide.util.Exceptions]
An error occurred during parsing of '[...]'. Please report a bug against java/source and attach dump file '[...]'.
An error occurred during parsing of '[...]'. Please report a bug against java/source and attach dump file '[...]'.
Caused: java.lang.AssertionError
	at jdk.compiler/com.sun.tools.javac.util.Assert.error(Assert.java:155)
	at jdk.compiler/com.sun.tools.javac.util.Assert.check(Assert.java:46)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$AliveAnalyzer.clearPendingExits(Flow.java:606)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$AliveAnalyzer.visitMethodDef(Flow.java:597)
	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMethodDecl.accept(JCTree.java:967)
	at jdk.compiler/com.sun.tools.javac.tree.TreeScanner.scan(TreeScanner.java:50)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$BaseAnalyzer.scan(Flow.java:445)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$AliveAnalyzer.visitClassDef(Flow.java:576)
	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:866)
	at jdk.compiler/com.sun.tools.javac.tree.TreeScanner.scan(TreeScanner.java:50)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$BaseAnalyzer.scan(Flow.java:445)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$AliveAnalyzer.visitClassDef(Flow.java:557)
	at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:866)
	at jdk.compiler/com.sun.tools.javac.tree.TreeScanner.scan(TreeScanner.java:50)
	at jdk.compiler/com.sun.tools.javac.comp.Flow$BaseAnalyzer.scan(Flow.java:445)
[...]

Issue #588, opened 2026-04-24
javac-related ClassCastException with JDK 26


anthonyvdotbe:
extension version: 25.1.0

extension configuration:

```plaintext
"jdk.advanced.disable.nbjavac": true,
"jdk.jdkhome": "C:\\dev\\jdk-26",
```

stack trace:

```plaintext
[Error - 9:26:27 AM] Request textDocument/semanticTokens/full failed.
  Message: Internal error.
  Code: -32603 
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.netbeans.modules.java.lsp.server.protocol.Server$ConsumeWithLookup$2.lambda$consume$0(Server.java:342)
	at org.netbeans.modules.openide.util.GlobalLookup.execute(GlobalLookup.java:45)
	at org.openide.util.lookup.Lookups.executeWith(Lookups.java:287)
	at org.netbeans.modules.java.lsp.server.protocol.Server$ConsumeWithLookup$2.consume(Server.java:340)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:330)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1516)
[...]

Issue #446, opened 2025-08-07
Support javac of latest JDK


anthonyvdotbe:
To reproduce:

1. configure the extension like this:

```jsonc
    "jdk.advanced.disable.nbjavac": true,
    "jdk.jdkhome": "C:\\dev\\jdk-26",
    "jdk.verbose": true,
```

where jdk-26 is [build 9](https://jdk.java.net/26/)

2. open Java projects

Expected: everything works.

Actual: `No suggestions.` in the editor and plenty of errors in the output.

This is probably the root cause:

```plaintext
java.lang.IllegalStateException: java.lang.NoSuchMethodError: 'void com.sun.tools.javac.parser.JavacParser$AbstractEndPosTable.<init>(com.sun.tools.javac.parser.JavacParser)'
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.parse(JavacTaskImpl.java:252)
	at org.netbeans.modules.java.source.parsing.JavacParser.moveToPhase(JavacParser.java:661)
	at org.netbeans.modules.java.source.parsing.CompilationInfoImpl.toPhase(CompilationInfoImpl.java:426)
	at org.netbeans.api.java.source.CompilationController.toPhase(CompilationController.java:90)
	at org.netbeans.modules.java.source.JavaSourceUtilImpl$1.parseModule(JavaSourceUtilImpl.java:304)
	at org.netbeans.modules.java.preprocessorbridge.api.ModuleUtilities.parseModule(ModuleUtilities.java:80)
	at org.netbeans.modules.java.api.common.classpath.ModuleClassPaths$ModuleInfoClassPathImplementation.getResources(ModuleClassPaths.java:844)
	at org.netbeans.api.java.classpath.ClassPath.entries(ClassPath.java:363)
	at org.netbeans.spi.java.classpath.support.MuxClassPathImplementation.getResources(MuxClassPathImplementation.java:90)
	at org.netbeans.api.java.classpath.ClassPath.entries(ClassPath.java:363)
	at org.netbeans.api.java.classpath.ClassPath.toString(ClassPath.java:769)
	at org.netbeans.api.java.classpath.ClassPath.toString(ClassPath.java:838)
	at org.netbeans.api.java.classpath.ClassPath.toString(ClassPath.java:848)
	at org.netbeans.modules.parsing.impl.indexing.LogContext.addPaths(LogContext.java:892)
[...]

PR #592, opened 2026-04-24
Updated README with the known issue regarding disable nb-javac setting

sid-srini:
Known issue refers to the root cause described in #446, #588, #589 etc.


Achal1607 (approved):
LGTM. Thanks @sid-srini

PR #593, opened 2026-04-27
Fixed SBOM generation failure on older node versions

sid-srini:
Pinned an earlier version of cdxgen to avoid transitive dependencies failing on earlier node versions.


Achal1607 (dismissed):
LGTM. Thanks @sid-srini

Achal1607 (dismissed):

Achal1607 (dismissed):

Achal1607 (approved):

PR #598, opened 2026-05-07
Amended README and Changelog for v26.0.0 features

sid-srini:
- Fixed merge issue in README causing reordering of the Maven user settings and VS Code variable substitution sections.
- Added a note about working around VS Code variable substitution in settings when using literal string that looks like a variable reference.
- Commented out shields badges for VS Marketplace that have been retired (Refer: vsmarketplace issue 1776).
- Added Changelog entries for new static content changes in v26.0.0


Achal1607 (approved):
LGTM. Thanks @sid-srini

PR #606, opened 2026-06-05
Fixed jdkHome config value handler fallback to env vars

sid-srini:
Fixed `handler.ts:jdkHomeValueHandler()` to have the promise return the fallback env vars `JDK_HOME` or `JAVA_HOME` when no config value is defined.
- The config getter returns the default value i.e. `null`, in such a case.
- Thus, the `defaultValue` parameter cannot be used, since the vscode config getter returns the default value defined in package.json, causing the value to not be `undefined`.
- So, a `then()` clause on the promise is needed to check for `null` or `undefined` value, and fallback to the env vars, if so.


Achal1607 (approved):
Hey, @sid-srini thanks for identifying it. LGTM.

PR #607, opened 2026-06-05
Added basic type checks in java.lsp.server config change listeners

sid-srini:
Backports netbeans PR 9369.
- Added checks for null and required JsonElement type in the registered config change listeners in WorkspaceService.
- Included a check for blank input defaultPlatformOverride in AbstractJavaPlatformProviderOverride to avoid expensive initialization in such cases.

This fixes issues like "IOException: Invalid J2SE platform" when `jdk.project.jdkhome` is set to `""` in settings.json.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -61,6 +61,7 @@
             patches/9263-draft.diff
             patches/9282-draft.diff
             patches/9290.diff
+            patches/9369.diff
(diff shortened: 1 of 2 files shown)


Achal1607 (approved):
LGTM. Thanks @sid-srini

PR #638, opened 2026-08-25
Reinstated the config to disable nbjavac

sid-srini:
javac from newer JDK EA versions (JDK 27+) may be used directly with the NB31 update.


singh-akhilesh (approved):
LGTM

